PR #1, opened 2016-11-08
init setup

RNAer:
@amnona, let's start populating it!

### DIFF
--- a/calour/study.py
+++ b/calour/study.py
@@ -34,14 +34,14 @@ class Study:
     feature_metadata : ``pandas.DataFrame``
     '''
     def __init__(self, data, sample_metadata, feature_metadata=None,
-                 description='', sparse=False):
+                 description='', sparse=True):
         self.data = data
         self.sample_metadata = sample_metadata
         self.feature_metadata = feature_metadata
         self.description = description
 
     @staticmethod
-    def _read_biom(fp, transpose=True, sparse=False):
+    def _read_biom(fp, transpose=True, sparse=True):
         '''Read in a biom table file.
 
         Parameters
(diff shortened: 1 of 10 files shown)


calour/study.py
```diff
@@ -0,0 +1,138 @@
...
+            column while sklearn and other packages require samples in
+            row. So you should transpose the data table.
+        '''
+        table = biom.load_table(fp)
+        sid = table.ids(axis='sample')
+        oid = table.ids(axis='observation')
+        if sparse:
+            data = scipy.sparse.csr_matrix(table.matrix_data)
+        else:
+            data = table.matrix_data.todense().A
```
wasade: This returns a matrix type not an array, probably want the `toarray` method

Are you _sure_ it makes sense to avoid a sparse matrix?

amnona: regarding avoiding sparse, some methods are faster on dense arrays, so sometimes useful to store as dense to begin with.
but agree can change default to sparse

wasade:
If defaulting to sparse, then it seems like quite a bit of replication of `biom.Table`? 

Note: in BIOM, we originally supported both dense and sparse, only to realize that the support for dense made the code much more complex and the subsequent benefit in practice was essentially non existent

amnona:
While making the code more complex, i think there are a lot of use cases for analysis where we need a dense matrix, and converting each time before running can slow down the processing?

wasade:
meh...? This is with a much denser matrix that we'd normally be operating on (50%)

```python
In [9]: x = np.random.randint(0, 2, (10000, 1000))

In [10]: y = csr_matrix(x)

In [11]: %timeit y.toarray()
10 loops, best of 3: 62.1 ms per loop
```

PR #4, opened 2016-11-12
rename some files/dirs

amnona:


calour/__init__.py
```diff
@@ -8,3 +8,8 @@
 
 __credits__ = "https://github.com/biocore/calour/graphs/contributors"
 __version__ = "0.1.0.dev0"
+
+from calour.experiment import *
+from calour.io import *
+from calour.filtering import *
+from calour.sorting import *
```
RNAer: I don't think we need these imports

calour/experiment.py
```diff
@@ -15,6 +15,8 @@
 import scipy
 import biom
 
+import calour as ca
+
```
RNAer: could you remove the changes of `experiment.py` for now? It will conflict with my changes.

calour/filtering.py
```diff
@@ -1,6 +1,8 @@
 # calour functions for filtering samples/observations
 # functions should call reorder_samples() / reorder_obs()
 
+import calour as ca
+
```
RNAer: use `from .experiment import Experiment` instead.

calour/__init__.py
amnona: i think this is like in numpy
we import all functions from all modules so they are all available under calour.XXX
this is so the user does not need to remember in which module is function is located
so when you use calour, you can just do import calour and then use whatever function you want...

calour/experiment.py
amnona: this made calour work (so i can use the read and test the functions i'm writing...)
maybe you/i can merge the PRs and resolve the conflicts?
what can we do?

calour/filtering.py
amnona: i want to have all the calour functions under one place (like importing numpy...) why not do it?

calour/__init__.py
RNAer: with dynamically import functions into class methods, we only need to import the class to access them.

calour/experiment.py
RNAer: it is probably easier to just revert the changes for the time being. saw you just merged my PR. maybe update based on that?

calour/__init__.py
amnona: right :)
is there an elegant way to import all functions into the class (i.e. without explicitly writing each function name?)

RNAer:
I already did it in the PR you merged :-)

calour/__init__.py
wasade: please do not do `import *`, it greatly complicates understanding codebases.

calour/experiment.py
```diff
@@ -44,19 +46,33 @@ def __init__(self, data, sample_metadata, feature_metadata=None,
...
 
+        # flag if data array is sparse (True) or dense (False)
+        self.sparse = sparse
+
     def __repr__(self):
         '''
         print the information about the experiment
         should have number of samples, observations, first 3 sequences and first 3 samples?
         '''
+        return('Experiment %s with %d samples, %d features' % (self.description,self.data.shape[0],self.data.shape[1]))
```
wasade: please make sure flake8 is being executed by travis

calour/normalization.py
```diff
@@ -0,0 +1,45 @@
...
+    -------
+    newexp : Experiment
+        the normalized experiment
+    '''
+    if not inplace:
+        exp = deepcopy(exp)
+
+    normfactor = reads / np.sum(exp.data, axis=axis)
+    if axis==0:
+        exp.data = exp.data * normfactor[:, None]
```
wasade: both branches are the same?

wasade:
Are there tests?

PR #40, opened 2017-07-26
add subsample function

RNAer:
@amnona @mortonjt , could any of you review this PR?

### DIFF
--- a/calour/transforming.py
+++ b/calour/transforming.py
@@ -2,6 +2,8 @@
 transforming (:mod:`calour.transforming`)
 =========================================
 
+.. warning:: Some of the functions require dense matrix and thus will change the sparse matrix to dense matrix.
+
 .. currentmodule:: calour.transforming
 
 Functions
@@ -18,6 +20,7 @@ Functions
    log_n
    transform
    center_log
+   subsample_count
 '''
 
 # ----------------------------------------------------------------------------
@@ -37,6 +40,8 @@ from sklearn import preprocessing
 
 from .experiment import Experiment
 from skbio.stats.composition import clr, centralize
+from skbio.stats import subsample_counts
+
 
 logger = getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


calour/transforming.py
```diff
@@ -345,3 +351,50 @@ def method(x): return x + delta
...
+
+    """
+    if inplace:
+        newexp = exp
+    else:
+        newexp = deepcopy(exp)
+    if newexp.sparse:
+        newexp.sparse = False
+    # subsample_counts() require int as input
+    newexp.data = newexp.data.astype(int)
```
amnona: maybe worth testing it is actually integers so if not, raise an error that it only works on non-normalized data?
alternatively, can check exp.exp_metadata['normalized'] (we set it if we run exp.normalize)

RNAer:
I can check exp.exp_metadata['normalized'], but it looks like biom read all
tables as float (that's why I did the int conversion), what do u suggest?

amnona:
i think worth checking  exp.exp_metadata['normalized'] (to see if exists and non zero) and in that case raise data has been normalized.
Alternatively, can test if all (float) values are integers (i.e. x.0), but i think less good than checking 'normalized' since can imagine crazy case where after normalization all numbers are integers by chance :)
if we convert to int, maybe we should convert back to float after subsampling? (just to make sure biom table is always float)? or maybe not needed?

calour/transforming.py
```diff
@@ -345,3 +351,50 @@ def method(x): return x + delta
...
+    drops = []
+    for row in range(newexp.data.shape[0]):
+        try:
+            newexp.data[row, :] = subsample_counts(newexp.data[row, :], n=total, replace=replace)
+        except ValueError:
+            drops.append(row)
```
mortonjt: Couple of minor questions
1. What's the try for?
2. Can we just run an `apply_along_axis` function?

RNAer: if replace is False, and a sample have smaller count than the total, this sample should be dropped. that's the `try` for. That's also why I am not sure `apply_along_axis` works in this situation.

RNAer:
@amnona @mortonjt , updated with normalization check and more explaining documentation for the code

calour/transforming.py
mortonjt: 👍 ok that makes sense.  `apply_along_axis` probably won't work then.  Do you want to add a comment about why this occurs for future reference?

RNAer:
yep. I did

PR #49, opened 2017-09-14
add filter_sample_categories and allow slice in __getitem__

RNAer:
delete extra data/notebooks plus other minor changes


amnona:
unit testing is failing (one test has a problem in test_filtering...)

calour/analysis.py
```diff
@@ -65,13 +65,14 @@ def correlation(exp, field, method='spearman', nonzero=False, transform=None, nu
         the desired FDR control level
     numperm : int
         number of permutations to perform
+    fdr_method : str
+        method to compute FDR. Allowed method include "", ""
```
amnona: the FDR procedure to determine significant bacteria
        'dsfdr' : discrete FDR method
        'bhfdr' : Benjamini-Hochberg FDR method
        'byfdr' : Benjamini-Yekutielli FDR method
        'filterBH' : Benjamini-Hochberg FDR method with filtering (automatic removal of features which cannot reach significance level)

calour/filtering.py
```diff
@@ -97,6 +98,44 @@ def downsample(exp, field, axis=0, num_keep=None, inplace=False):
...
+    Examples
+    --------
+
+    Parameters
+    ----------
+    field : str
+        The name of the column in samples metadata table. This column
+        should has categorical values
+    min_samples : int (optional)
+        Filter away the samples with a value in the given column if its sample count is
```
amnona: Maybe default should be 0? so no dropping with min_samples unless specified?

amnona:
added a few silly comments in the code, but looks good :)

RNAer:
nope. if 0, then you don't need to run this function at all.

PR #51, opened 2017-09-15
Qt5labels

amnona:
Fix sample/feature labels in qt5 gui so text  will be visible on lower res. display


RNAer:
could you paste snapshots here to show it before and after the changes? and add comments on the changed code. otherwise, no one will remember after a few months, making it difficult to maintain.

amnona:
before the fix:
image

after:
image

PR #53, opened 2017-09-16
add conda recipes in this repo

RNAer:


amnona:
cool!
2 questions:
1. do we want to make it conda installable? Will it make updating versions more complicated every time we fix something? how complicated is it to update the conda install?
2. what about the related modules (ezcalour, calour-dbbact etc.)? Should we add to the default conda install?

RNAer:
1. we will make every release installable. probably delete the build of
master branch?

2. we can make them conda installable too.

PR #62, opened 2017-09-30
 refactor read() to add kwargs for reading metadata and add logging unit tests

RNAer:
simplify test1/test2 data sets. @amnona , could you take a look?


amnona:
the md5 unit test is failing since you created a new dataset - put in the new md5 instead :)

calour/_testing.py
```diff
@@ -17,6 +18,9 @@
 
 class Tests(TestCase):
     def setUp(self):
+        # disable logging
+        logging.disable(logging.CRITICAL)
```
amnona: why disable logging.critical ? what does this line do?

calour/filtering.py
```diff
@@ -326,7 +328,7 @@ def _mean_abundance(data, axis, cutoff=0.01, strict=False):
     return res
 
 
-def _prevalence(x, cutoff=1/10000, fraction=0.5):
+def _prevalence(x, cutoff=1/10000, fraction=0.1):
```
amnona: i like the default to be 0.5, so it keeps features present in at least 0.5 the samples.... but i see the logic in your suggestion... maybe remove the default and make it mandatory?

calour/io.py
```diff
@@ -286,8 +261,33 @@ def read_open_ms(data_file, sample_metadata_file=None, gnps_file=None, feature_m
     return exp
 
 
+def _read_metadata(ids, f, kwargs):
+    # load the sample/feature metadata file
```
amnona: i think we should add the parameters/returns docstring even though it's an internal function (so we remember how to use it :)   )

calour/io.py
```diff
@@ -286,8 +261,33 @@ def read_open_ms(data_file, sample_metadata_file=None, gnps_file=None, feature_m
...
+        metadata = metadata.loc[ids, ]
+    else:
+        metadata = pd.DataFrame(index=ids)
+    return metadata
+
+
 def read(data_file, sample_metadata_file=None, feature_metadata_file=None,
-         description='', drop=False, sparse=True, data_file_type='biom', encoding=None,
+         description='', sparse=True, data_file_type='biom',
+         sample_metadata_kwargs=None, feature_metadata_kwargs=None,
```
amnona: what about encoding? sometimes i handle mapping files that fail with utf (so i load them with encoding='latin8'. It is really useful.

calour/io.py
```diff
@@ -316,102 +314,68 @@ def read(data_file, sample_metadata_file=None, feature_metadata_file=None,
         'openms' : an OpenMS bucket table csv (rows are feature, columns are samples)
         'openms_transpose' an OpenMS bucket table csv (columns are feature, rows are samples)
         'qiime2' : a qiime2 biom table artifact (need to have qiime2 installed)
-    encoding : str or None (optional)
-        encoder for the metadata files. None (default) to use
-        pandas default encoder, str to specify encoder name (see
-        pandas.read_table() documentation)
+    sample_metadata_kwargs, feature_metadata_kwargs : dict or None (optional)
+        keyword arguments passing to :function:`pandas.read_table` when reading sample metadata
+        or feature metadata
```
amnona: maybe add here example for encoding. i.e. such as:
for example use 'encoding:"latin-8"'

calour/io.py
```diff
@@ -286,8 +261,33 @@ def read_open_ms(data_file, sample_metadata_file=None, gnps_file=None, feature_m
...
+        # reorder the id in metadata to align with biom
+        metadata = metadata.loc[ids, ]
+    else:
+        metadata = pd.DataFrame(index=ids)
+    return metadata
+
+
 def read(data_file, sample_metadata_file=None, feature_metadata_file=None,
-         description='', drop=False, sparse=True, data_file_type='biom', encoding=None,
+         description='', sparse=True, data_file_type='biom',
```
amnona: what about the 'drop' parameter?
we used if you want to keep samples not in the mapping file or alternatively drop them. 
Or do you want users to filter it later with a filter_samples command?

calour/io.py
```diff
@@ -316,102 +314,68 @@ def read(data_file, sample_metadata_file=None, feature_metadata_file=None,
...
-        feature_metadata['id'] = oid
-    if data_file_type == 'biom' and md is not None:
-        # combine it with the metadata
-        feature_metadata = pd.concat([feature_metadata, md], axis=1)
+    sample_metadata = _read_metadata(sid, sample_metadata_file, sample_metadata_kwargs)
+    feature_metadata = _read_metadata(fid, feature_metadata_file, feature_metadata_kwargs)
+
+    if fmd is not None:
+        # combine it with the feature metadata
+        feature_metadata = pd.concat([feature_metadata, fmd], axis=1)
```
amnona: I like to have info about the difference between the mapping and biom table samples (i.e. how many were present in one but not the other). It lets you get some quick info about the success of sequencing and errors in the mapping file.
This part was deleted.... maybe we can add something for that end?

amnona:
Also, just for general knowledge, why did you change the test dataset? to make it simpler?

amnona:
@RNAer looks nice, a few comments/questions :)

calour/_testing.py
RNAer: to suppress the log msg in all the tests. not critical, just make the screen cleaner when running tests.

calour/io.py
RNAer: ok. will add that

calour/io.py
RNAer: will add

calour/io.py
RNAer: this part is in the _read_metadata()

calour/io.py
RNAer: i think we might wanna drop samples w/o metadata but don't wanna drop features w/o metadata. so use filter to drop these rows or columns after reading it. we set `drop=False` before anyway.

calour/_testing.py
amnona: i like to have the log messages since it helps find out the bug when tests fail - but whatever you prefer.

calour/filtering.py
amnona: let's make "fraction" mandatory (can move it before "cutoff")?

calour/io.py
```diff
@@ -286,8 +261,48 @@ def read_open_ms(data_file, sample_metadata_file=None, gnps_file=None, feature_m
     return exp
 
 
+def _read_metadata(ids, f, kwargs):
+    '''read metadata table
+
+    Parameters
+    ----------
+    ids : list like of str
+        ids from data table
```
amnona: can you add explanation what ids is used for (i.e. reorder and keep only data about ids)

calour/io.py
```diff
@@ -286,8 +261,48 @@ def read_open_ms(data_file, sample_metadata_file=None, gnps_file=None, feature_m
...
+        metadata = pd.read_table(f, **kwargs)
+        mid, ids2 = set(metadata.index), set(ids)
+        diff = mid - ids2
+        if diff:
+            logger.warning('These have metadata but do not have data - dropped: %r' % diff)
+        diff = ids2 - mid
+        if diff:
+            logger.warning('These have data but do not have metadata: %r' % diff)
+        # reorder the id in metadata to align with biom
+        metadata = metadata.loc[ids, ]
```
amnona: what about if some samples/features are not in the metadata file? i think we should add them to the dataframe with "NA" (rather than just showing a warning). Otherwise it will cause bugs downstream.

RNAer: yes. those are added as NA. it is tested in modified tests

calour/io.py
RNAer: this explanation shouldn't be here? remember this is private function

calour/filtering.py
RNAer: i disagree. the previous 0.5 is too aggressive. what reason to make it mandatory?

calour/_testing.py
RNAer: the unit tests are already very informative in printing error msg...

calour/filtering.py
amnona: since we see people expect different values as default :)

PR #65, opened 2017-10-13
update heatmap

RNAer:
to plot continuous sample tick lables (eg ph)

For small sample / feature size, you will likely have white lines everywhere:

image

### DIFF
--- a/calour/heatmap/heatmap.py
+++ b/calour/heatmap/heatmap.py
@@ -87,11 +87,14 @@ def _truncate_middle(x, length=16):
 
 
 def _set_axis_ticks(ax, which, ticklabels, tickmax, n, kwargs, ticklabel_len):
-    ''''''
-    try:
-        ticks = _transition_index(ticklabels)
-    except KeyError:
-        raise ValueError('Sample field %r not in sample metadata' % sample_field)
+    '''Plot tick and ticklabels of x or y axis.
+
+    Parameters mean axes, x or y axis, list of tick labels, max number
+    of ticks, the upper bound of xlim/ylim, dict passing as text
+    property, the length of each tick label.
+
+    '''
+    ticks = _transition_index(ticklabels)
     tick_pos, tick_val = zip(*ticks)
 
     axis = getattr(ax, which + 'axis')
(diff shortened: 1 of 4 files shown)


calour/heatmap/heatmap.py
```diff
@@ -238,49 +283,21 @@ def heatmap(exp: Experiment, sample_field=None, feature_field=None,
...
-            else:
-                return ''
-        # set the maximal number of feature labels
-        ax.yaxis.set_major_formatter(mpl.ticker.FuncFormatter(format_fn))
-        ax.yaxis.set_major_locator(mpl.ticker.MaxNLocator(max_yticks, integer=True))
-        if yticklabel_kwargs is None:
-            yticklabel_kwargs = {'rotation': 45, 'ha': 'right', 'size': 6}
-        for t in ax.get_yticklabels():
-            t.set(**yticklabel_kwargs)
+        _set_axis_ticks(ax, 'y', ticklabels, yticks_max, numrows, yticklabel_kwargs, xticklabel_len)
```
amnona: isn't it supposed to be yticklabel_len and not xticklabel_len?

calour/heatmap/heatmap.py
```diff
@@ -86,10 +86,55 @@ def _truncate_middle(x, length=16):
         return ['%s..%s' % (i[:mid], i[-mid:]) if len(i) > length else i for i in y]
 
 
+def _set_axis_ticks(ax, which, ticklabels, tickmax, n, kwargs, ticklabel_len):
+    ''''''
```
amnona: i think it's worth having a (small) description & parameter list for this function? otherwise in a few years we will forget what it does and how :)
i.e. what is 'which'

calour/heatmap/heatmap.py
```diff
@@ -86,10 +86,55 @@ def _truncate_middle(x, length=16):
         return ['%s..%s' % (i[:mid], i[-mid:]) if len(i) > length else i for i in y]
 
 
+def _set_axis_ticks(ax, which, ticklabels, tickmax, n, kwargs, ticklabel_len):
```
amnona: why is kwargs not **kwargs? and why is it before ticklabel_len?
(probably my lack of knowledge in python...)

calour/heatmap/heatmap.py
RNAer: typo. fixed

calour/heatmap/heatmap.py
RNAer: added

calour/heatmap/heatmap.py
RNAer: it is you lack of python knowledge :). If you don't need to expand the dict into key word arguments, you shouldn't use `**`

PR #69, opened 2017-10-17
Limit scroll

amnona:
fix scrolling with arrows gets out of data bounds


calour/calour.config
```diff
@@ -1,8 +1,10 @@
 [dbbact]
 module_name = dbbact_calour.dbbact
 class_name = DBBact
-username = pitapita
+username = amnon
 password = pitapitapita
+skip_msg_study_info = yes
+skip_msg_annotation_info = yes
```
RNAer: what are these attributes?

RNAer: put on comments?

calour/heatmap/plotgui.py
```diff
@@ -328,12 +328,20 @@ def key_press_callback(self, event):
...
         elif event.key == 'up':
+            if ylim_upper + y_offset < -0.5:
+                y_offset = -0.5 - ylim_upper
             ax.set_ylim(ylim_lower + y_offset, ylim_upper + y_offset)
         elif event.key == 'left':
+            if xlim_lower - x_offset < -0.5:
+                x_offset = xlim_lower + 0.5
             ax.set_xlim(xlim_lower - x_offset, xlim_upper - x_offset)
         elif event.key == 'right':
+            if xlim_upper + x_offset > self.exp.data.shape[0] - 0.5:
```
RNAer: minor, but use `right_end = self.exp.data.shape[0] - 0.5` to reuse the value below?

same for `down` key

PR #71, opened 2017-11-15
Calour pc installation instructions

amnona:
+ link from main README.md
Also:
- Update read function when feature-metadata and biom table have same column name (i.e. taxonomy)
- Fix small problem in dsFDR in stdmean method when std is 0


amnona:
@RNAer can you merge?

installation-pc.md
```diff
@@ -0,0 +1,134 @@
...
+go to [http://landinghub.visualstudio.com/visual-cpp-build-tools](http://landinghub.visualstudio.com/visual-cpp-build-tools), download "Build tools 2015", and run the installer (You can select all default options).
+
+Then retry:
+```
+pip install biom-format
+```
+
+Install scikit-bio dependency:
+```
+conda install -c biocore biom-format scikit-bio
```
RNAer: why `biom-format` again?

calour/dsfdr.py
```diff
@@ -54,7 +54,9 @@ def stdmeandiff(data, labels):
     mean1 = np.mean(data[:, labels == 1], axis=1)
     sd0 = np.std(data[:, labels == 0], axis=1, ddof=1)
     sd1 = np.std(data[:, labels == 1], axis=1, ddof=1)
-    tstat = (mean1 - mean0) / (sd1 + sd0)
+    sdsum = sd0 + sd1
+    sdsum[sdsum == 0] = 1
```
RNAer: could you add comment on why update them to 1?

calour/io.py
```diff
@@ -384,6 +384,13 @@ def read(data_file, sample_metadata_file=None, feature_metadata_file=None,
     # self.feature_metadata['_calour_original_abundance'] = self.data.sum(axis=0)
 
     if fmd is not None:
+        # rename columns in biom table if exist in feature metadata file
+        renames = {}
+        for ccol in fmd.columns:
+            if ccol in feature_metadata.columns:
+                renames[ccol] = ccol+'_biom'
+            if len(renames) > 0:
```
RNAer: not big deal, but `if renames` is enough

calour/dsfdr.py
amnona: added comment in text.
mean/std is not defined when values are constant on each group
1 is arbitrary... but as good as any (could use small epsilon instead, but i think it inflates this strange result...)

calour/io.py
amnona: cool. fixed.

installation-pc.md
amnona: oops. nice catch :)
fixed to pip install since this is windows

amnona:
addressed comments :)
(also added a fix for ezcalour and qt5)
@RNAer - ready for merge?

calour/heatmap/plotgui_qt5.py
```diff
@@ -55,13 +61,8 @@ def __call__(self):
...
             # run the event loop
             self.app.exec_()
-        finally:
-            # clean up when the qt app is closed
-            if self.app_window in self.app.references:
-                logger.debug('removing window from app window list')
-                self.app.references.remove(self.app_window)
-            else:
-                logger.debug('window not in app window list. Not removed')
+        except:
```
RNAer: are you able to provide exception type? it is very dangerous without specifying exception type.

why removing the cleanup code after closing qt gui? Oh, I see. it's moved to `fileQuit` which is called by `closeEvent`. where is `closeEvent` used? could you comment it inside the `closeEvent` function?

btw, is this change for Windows OS?

calour/io.py
```diff
@@ -389,7 +389,7 @@ def read(data_file, sample_metadata_file=None, feature_metadata_file=None,
         for ccol in fmd.columns:
             if ccol in feature_metadata.columns:
                 renames[ccol] = ccol+'_biom'
-            if len(renames) > 0:
+            if renames > 0:
```
RNAer: this is wrong. pls drop `> 0`...

amnona: oopsi
fixed :)

calour/heatmap/plotgui_qt5.py
amnona: closeEvent is the default class method qt5 calls when a window is closed.
The change is for ezcalour.
I used a general except since we don't know what error we'll get - we try to let the program be a little robust and not fail and stop... but maybe we should let it fail if opening the window fails?
added a comments to explain better :)

amnona:
ok. try #2 :)
@RNAer - how is it now?

calour/heatmap/plotgui_qt5.py
RNAer: you can fail it once and see what type of exception it is?

amnona:
ok, removed the except :)
it was a general just in case error catching, but fine to just fail...
@RNAer merge?

PR #81, opened 2018-01-14
Diff group

amnona:
Add _calour_diff_abundance_group to each feature (i.e. name of the group where it is higher)


calour/plotting.py
```diff
@@ -107,12 +112,12 @@ def plot_enrichment(exp: Experiment, enriched, max_show=10, max_len=40, ax=None)
     positive = np.min([np.sum(enriched['odif'].values > 0), max_show[0]])
     negative = np.min([np.sum(enriched['odif'].values < 0), max_show[1]])
     if negative + positive == 0:
-        print('No significantly enriched categories found')
+        logger.warn('No significantly enriched categories found')
```
RNAer: i think it's `logger.warning`? if i recall correctly

PR #83, opened 2018-02-20
clean up documentation

RNAer:
1. drop numpydoc (and also the verbose cross-reference) in favor of napoleon plugin

2. merge README.md content to INSTALL.md


doc/source/getting_started.rst
```diff
@@ -14,37 +14,8 @@ Calour mostly handles **Experiment** data. An Experiment is made of **Samples**,
...
-
-In order to run the jupyter notebook server, change directory to the directory where your analysis notebook is located and then type:
-
-```jupyter notebook```
-
-and within the notebook, just import the Calour module using:
-
-```import calour as ca```
-
-For a generic Calour microbiome analysis notebook that contains the standard analysis workflow, you can download the `simple_microbiome_analysis.ipynb <https://raw.githubusercontent.com/biocore/calour/master/notebooks/demo.ipynb>`_.
```
amnona: Did it move to a different file? or just deleted? If the latter, i think we should keep something like this (from my experience with microbiome analysis users, some will get stuck on how to run calour and use it for the first time... Maybe should put in a different file/section...?

amnona:
cool,
byebye numpydoc.
Did we change docrep to handle the new section name?
also, i think we should keep the more elaborate instructions for how to run calour (so it will be step by step... including how to load the sample notebook...)
but besides that very cool.

doc/source/getting_started.rst
RNAer: deleted. I documented that your tutorial notebooks is available to download to run. if the case, could you put this notebook as the first tutorial notebook in the tutorial section later? BTW, I think you should re-order the notebooks in that section later - some notebooks should probably be put before others.

RNAer:
yes, changed. ready for merge?

PR #84, opened 2018-02-22
read exported gnps ms tables

amnona:
read exported gnps bucket table, and improve mass-spec mz/rt and gnps database link.
Also bug fixes for sample/feature metadata in interactive plot in jupyter notebook.
and fix html link for database info in jupyter notebook interactive plot


calour/database.py
```diff
@@ -230,6 +230,25 @@ def get_seq_annotation_strings(self, sequence):
         logger.debug('Generic function for get_annotation_strings')
         return []
 
+    def get_annotation_website(self, annotation):
+        '''Get the databasewebsite address of information about the annotation.
```
RNAer: missing a space

calour/heatmap/heatmap.py
```diff
@@ -164,8 +164,11 @@ def heatmap(exp: Experiment, sample_field=None, feature_field=None,
     ----------
     sample_field : str or None, optional
         The field of sample metadata to display on the x-axis or None (default) to not show x axis.
-    feature_field : str or None, optional
-        The field of feature meadata to display on the y-axis or None (default) to not show y axis.
+    feature_field : str or None or False (optional)
+        The field of feature meadata to display on the y-axis.
+        False to use the default feature field (set by experiment type)
+        None (default) to not show y axis.
+    xticklabel_kwargs, yticklabel_kwargs : dict or None (optional)
```
RNAer: pls don't change the above line. the price we pay to drop numpydoc is that we can have multiple func parameters on the same line - that's why I changed in my last PR

calour/heatmap/heatmap.py
```diff
@@ -164,8 +164,11 @@ def heatmap(exp: Experiment, sample_field=None, feature_field=None,
     ----------
     sample_field : str or None, optional
         The field of sample metadata to display on the x-axis or None (default) to not show x axis.
-    feature_field : str or None, optional
-        The field of feature meadata to display on the y-axis or None (default) to not show y axis.
+    feature_field : str or None or False (optional)
```
RNAer: I still don't get why we need to set a default feature field? allowing both `None` and `False` is confusing to users.

calour/heatmap/plotgui_qt5.py
```diff
@@ -81,10 +81,10 @@ def _update_info_labels(self, sid, fid, abd):
         self.app_window.w_sid.setText(str(sid))
         sample_field = str(self.app_window.w_sfield.currentText())
         self.app_window.w_sfield_val.setText(
-            str(self.exp.sample_metadata[sample_field][self.current_select[0]]))
+            str(self.exp.sample_metadata[sample_field].iloc[self.current_select[0]]))
```
RNAer: why is this change?

calour/io.py
```diff
@@ -472,12 +477,82 @@ def read_open_ms(data_file, sample_metadata_file=None, gnps_file=None, feature_m
...
         exp.exp_metadata['_calour_metabolomics_gnps_table'] = gnps_data
+        # link each feature to the gnps ids based on MZ/RT
+        exp._prepare_gnps_ids()
         # add gnps names to the features
         exp._prepare_gnps()
 
     return exp
 
 
+def read_gnps_ms(data_file, sample_metadata_file=None, gnps_file=None, feature_metadata_file=None,
```
RNAer: add a small test for this read func?

calour/io.py
```diff
@@ -472,12 +477,82 @@ def read_open_ms(data_file, sample_metadata_file=None, gnps_file=None, feature_m
...
+        None (default) to not load
+    description : str or None, optional
+        Name of the experiment (for display purposes).
+        None (default) to assign file name
+    sparse : bool, optional
+        False (default) to store data as dense matrix (faster but more memory)
+        True to store as sparse (CSR)
+    normalize : int or None
+        normalize each sample to the specified reads. None to not normalize
+
```
RNAer: add docrep as in `read_amplicon`?

calour/io.py
```diff
@@ -472,12 +477,82 @@ def read_open_ms(data_file, sample_metadata_file=None, gnps_file=None, feature_m
...
+    '''Load a GNPS metabolomics experiment (exported from GNPS).
+    Parameters
+    ----------
+    data_file : str
+        name of the OpenMS bucket table CSV file.
+    sample_metadata_file : str or None, optional
+        None (default) to not load metadata per sample
+        str to specify name of sample mapping file (tsv)
+    gnps_file : str or None, optional
+        name of the gnps clusterinfosummarygroup_attributes_withIDs_arbitraryattributes/XXX.tsv file
```
RNAer: missing separations between words

calour/database.py
amnona: fixed

calour/heatmap/heatmap.py
amnona: The idea is that for ms data, we want to have ticks for the y-axis with the metabolite name by default, whereas for amplicon we don't want the y axis ticks by default.
I don't think it will be too confusing if explained in the doc? 
alternatively, we can remove the default ytick for ms, and have users supply it explicitly, but then less convenient for the naive mass-spec user?

calour/heatmap/heatmap.py
amnona: fixed.
byebye numpydoc :)

calour/heatmap/plotgui_qt5.py
amnona: otherwise, if the index is an int (such as happens in ms gnps table), it returns an error (looks for the number in the index instead of line number....)

calour/io.py
amnona: this is actually the gnps filename we need :)
see [here](https://bix-lab.ucsd.edu/display/Public/Cytoscape+2.8.X+Visualization+and+Analysis+Documentation)

calour/io.py
amnona: added :)

calour/io.py
amnona: nice catch :)

amnona:
ok. all comments addressed :)
ready for merge?

calour/heatmap/heatmap.py
RNAer: I prefer the latter, making the API behavior more consistent. I would argue for naive mass spec users, MZ yticklabel probably don't help them anyway? you can add a doc for feature_metadata, saying for ms exp, you can supply MZ.

I don't like this because 1) None and False is confusing even with doc (I often can't recall which is to disable the yticklabel); 2) it makes the classes have yet another ad hoc attribute of feature metadata.

Other than this, I am good to merge.

amnona:
ok, removed the default heatmap feature field
also catching the specific exception now for database not found
and fixed the unit test :)
should be good to go.

PR #86, opened 2018-02-28
add dropna function

RNAer:


amnona:
maybe worth changing the function name to filter_na instead of dropna (to be consistent with the rest of the filter_XXX function names)?
i still like filter_samples(values=None) better, but if you insist can keep as separate :)

PR #90, opened 2018-03-31
Database interface

amnona:
Add documentation for how to write a new database interface for calour
small doc fixed for the database.Database class to make it clearer


doc/source/database_interface.rst
```diff
@@ -0,0 +1,93 @@
+Creating database interfaces for Calour
+=======================================
+
+Calour is designed with an extendible database interface support. In this document we will describe the steps required to integrate your database into Calour.
```
RNAer: extensible

calour/database.py
```diff
@@ -383,6 +404,7 @@ def enrichment(self, exp, features, *args, **kwargs):
     def show_term_details(self, term, exp, features, *args, **kwargs):
         '''
         Show details about the specific term in the database and in what features it appears
+        This is an optional function for databases that support enrichment analysis (level L3).
```
RNAer: this does not looks like for enrichment...

calour/database.py
```diff
@@ -316,6 +333,9 @@ def delete_annotation(self, annotation_details):
 
     def remove_feature_from_annotation(self, features, annotation_details):
         '''remove a feature from the annotation in the database (if allowed)
+        This is an optional function for databases that support manual annotations (level L4).
```
RNAer: what is the diff between this func and `delete_annotation`?

doc/source/database_interface.rst
amnona: fixed  :)

calour/database.py
amnona: Improved the explanation. This function is used from the enrichment gui, so usually called from L3 integration.

calour/database.py
amnona: added explanation. since an annotation can contain many features, delete removes the annotation from all the features it contains, whereas remove_feature_from annotation keeps the annotation but removes the association with the specific feature.

amnona:
addressed comments. ready for merge :)

PR #91, opened 2018-04-03
new mass-spec read

amnona:
rewrite all the mass-spec read functions.
Now we support several mass-spec formats (mzmine2, openms, gnps exported biom table) using a unified read_ms() function.
We also enable changing specific parameters in the load, in order to support other formats (since it's all very fluid)


calour/io.py
```diff
@@ -270,6 +272,10 @@ def read(data_file, sample_metadata_file=None, feature_metadata_file=None,
         {'index_col': False}.
     cls : ``class``, optional
         what class object to read the data into (:class:`.Experiment` by default)
+    cut_sample_id_sep: None or str, optional
+        if not None, split each data file sampleID using this value (typically '_'), taking only the first part as the sampleID.
+    rows_are_samples: bool, optional
```
RNAer: how about changing it to `sample_in_row` or `transpose`?

calour/io.py
```diff
@@ -234,7 +236,7 @@ def _read_metadata(ids, f, kwargs):
 def read(data_file, sample_metadata_file=None, feature_metadata_file=None,
          description='', sparse=True, data_file_type='biom',
          sample_metadata_kwargs=None, feature_metadata_kwargs=None,
-         cls=Experiment, *, normalize):
+         cls=Experiment, cut_sample_id_sep=None, rows_are_samples=False, *, normalize):
```
RNAer: I don't like `cut_sample_id_sep` and I don't think we need it. It's only used in the MS read func, right? You can update IDs after read() in read_ms()

calour/io.py
```diff
@@ -302,10 +304,21 @@ def read(data_file, sample_metadata_file=None, feature_metadata_file=None,
...
+        # sid = sid.map(lambda x: x.split(cut_sample_id_sep)[0])
+        sid = [x.split(cut_sample_id_sep)[0] for x in sid]
+
     sample_metadata = _read_metadata(sid, sample_metadata_file, sample_metadata_kwargs)
     feature_metadata = _read_metadata(fid, feature_metadata_file, feature_metadata_kwargs)
 
-    # store the abund  per sample/feature before any procesing
+    # store the sample and feature ids also as a column (for sorting, etc.)
+    sample_metadata['_sample_id'] = sample_metadata.index.values
+    feature_metadata['_feature_id'] = feature_metadata.index.values
```
RNAer: we wanna minimize the changes in the read(). As `cut_sample_id_sep`, I don't think we should do this.

calour/io.py
```diff
@@ -334,6 +347,14 @@ def read(data_file, sample_metadata_file=None, feature_metadata_file=None,
     exp = cls(data, sample_metadata, feature_metadata,
               exp_metadata=exp_metadata, description=description, sparse=sparse)
 
+    # remove samples containing NaN in data table (can happen in _read_csv)
+    nan_samples = np.where(np.isnan(exp.data.sum(axis=1)))[0]
```
RNAer: similarly, we wanna minimize the changes in the read(). we shouldn't remove any if it's in the data file. For example, what if one wants to fill na with zero or impute with mean.

RNAer: if it is needed, then it should be a function in filtering module to drop na.

calour/io.py
```diff
@@ -393,36 +414,99 @@ def read_amplicon(data_file, sample_metadata_file=None,
 
 
 @ds.with_indent(4)
-def read_open_ms(data_file, sample_metadata_file=None, gnps_file=None, feature_metadata_file=None,
-                 description=None, sparse=False, rows_are_samples=False, mz_rt_sep=None, *, normalize, **kwargs):
-    '''Load an OpenMS metabolomics experiment.
+def read_ms(data_file, sample_metadata_file=None, gnps_file=None, feature_metadata_file=None,
```
RNAer: to be consistent, put `gnps_file` behind `feature_metadata_file`

calour/io.py
```diff
@@ -393,36 +414,99 @@ def read_amplicon(data_file, sample_metadata_file=None,
...
+            file_type='mzmine2', rows_are_samples=None, table_sep=None, mz_rt_sep='_', direct_ids=None,
```
RNAer: to be consistent with `read`, `file_type` -> `data_file_type`

calour/io.py
```diff
@@ -393,36 +414,99 @@ def read_amplicon(data_file, sample_metadata_file=None,
...
+            table is a tsv/json/hdf5 biom table, columns are samples.
+        'openms': load an openms output table
+            MZ and RT are obtained via the featureID (first column), which is assumed to be MZ_RT.
+            GNPS linking is indirect via the mz and rt threshold windows.
+            table is a csv table, columns are samples.
+        'gnps-ms2': load a gnps exported biom table
+            MZ and RT are obtained via the gnps_file if available, otherwise are NA
+            GNPS linking is direct via the first column (featureID).
+            table is a tsv/json/hdf5 biom table, columns are samples.
+    rows_are_samples: bool or None, optional
```
RNAer: mark

calour/io.py
```diff
@@ -393,36 +414,99 @@ def read_amplicon(data_file, sample_metadata_file=None,
...
-        name of the gnps clusterinfosummarygroup_attributes_withIDs_arbitraryattributes/XXX.tsv file
-        for use with the 'gnps' database in plot
+        name of the gnps clusterinfosummarygroup_attributes_withIDs_arbitraryattributes/XXX.tsv file, for use with the 'gnps' database.
+        This enables identification of the metabolites with known MS2 (for the interactive heatmap and sorting/filtering etc), as well as linking
+        to the gnps page for each metabolite (from the interactive heatmap - by double clicking on the metabolite database information).
+        Note: requires gnps-calour database interface module (see Calour installation instructions for details).
     feature_metadata_file : str or None (optional)
         Name of table containing additional metadata about each feature
         None (default) to not load
+    file_type: str, optional
```
RNAer: mark

calour/io.py
```diff
@@ -393,36 +414,99 @@ def read_amplicon(data_file, sample_metadata_file=None,
...
+            table is a csv table, columns are samples.
+        'gnps-ms2': load a gnps exported biom table
+            MZ and RT are obtained via the gnps_file if available, otherwise are NA
+            GNPS linking is direct via the first column (featureID).
+            table is a tsv/json/hdf5 biom table, columns are samples.
+    rows_are_samples: bool or None, optional
+        None (default) to use the file_type defaults.
+        False indicates columns are samples.
+        True indicates rows are samples.
+    table_sep: str or None, optional
```
RNAer: use `data` / `table` consistently

calour/io.py
```diff
@@ -393,36 +414,99 @@ def read_amplicon(data_file, sample_metadata_file=None,
...
+        True indicates the FeatureID field contains the feature MZ and RT (assuming it is of the form MZ_RT).
+        see also mz_rt_sep parameter.
+        False indicates featureID field is just a unique feature identifier.
+        Note that if direct_ids=True, get_mz_rt_from_feature_id must be None or False.
+    use_gnps_id_from_AllFiles: bool, optional
+        True (default) to link to the gnps_file using ID extracted from the AllFiles field. This is better since will work even if
+        different filtering was applied for the gnps and ms1 processing.
+        False to link to the gnps_file using the 'cluster index' field.
+        NOTE: if using the False value, you must make sure identical filtering to both pipelines.
+    cut_sample_id_sep: None or str, optional
```
RNAer: mark

calour/io.py
```diff
@@ -432,131 +516,105 @@ def read_open_ms(data_file, sample_metadata_file=None, gnps_file=None, feature_m
...
     --------
     read
     '''
-    logger.debug('Reading OpenMS data (OpenMS bucket table %s, map file %s)' % (data_file, sample_metadata_file))
-    if rows_are_samples:
-        data_file_type = 'openms_transpose'
+    if file_type == 'mzmine2':
+        if rows_are_samples is None:
+            rows_are_samples = False
+        if table_sep is None:
```
RNAer: you can set `','` instead of `None` as default

calour/io.py
```diff
@@ -432,131 +516,105 @@ def read_open_ms(data_file, sample_metadata_file=None, gnps_file=None, feature_m
...
 
     See Also
     --------
     read
     '''
-    logger.debug('Reading OpenMS data (OpenMS bucket table %s, map file %s)' % (data_file, sample_metadata_file))
-    if rows_are_samples:
-        data_file_type = 'openms_transpose'
+    if file_type == 'mzmine2':
+        if rows_are_samples is None:
```
RNAer: you can set `False` as default

calour/io.py
```diff
@@ -432,131 +516,105 @@ def read_open_ms(data_file, sample_metadata_file=None, gnps_file=None, feature_m
...
+            rows_are_samples = False
+        if table_sep is None:
+            table_sep = ','
+        if direct_ids is None:
+            direct_ids = True
+        data_file_type = 'csv'
+    elif file_type == 'biom':
+        if rows_are_samples is None:
+            rows_are_samples = False
+        if direct_ids is None:
```
RNAer: again. for immutable variable, you don't need to set None as default.

calour/io.py
amnona: i don't think we can do it not in the read() function, since the read function links the data and sample_metadata table indices (_read_metadata() line 231, called from read() line 318), and if the indices are not the same (i.e. sample ids haven't been cut, it will not know how to link, and will fail.

calour/io.py
amnona: I agree in general we should try to minimize the read() changes.
While we could write the read_ms() without these new fields, they are useful for regular analysis (when you want to sort your samples by their id). So i think while this change is not related to read_ms, it is more elegant this way (where we can treat the featureid and sampleid as regular metadata fields as well). And this does not affect any backwards compatibility (just adds 2 new fields in the exp metadata), so should not break anything.

calour/io.py
amnona: added in filtering.py
however, we need to call it before the normalization in read(), so i don't know how to enable users to read data with nans (since normalize fails with nans).
The problem is mzmine2 data contains a comma at the end of each line, which leads to a sample with nans in all features when reading using pandas.read_csv()
therefore, we need to remove them in order for normalization in the read() to work...
options are:
1. always filter the nans (current implementation - now via calling the filter_data_na() from filtering.py in read()). problem is it will do it for all data, so will remove nans from all tables.
2. add a flag to say whether to filter nans or not for the read() function, and then filter/not filter according to it, and set it to true when reading mzmine2 data
3. fix the _read_csv() function to specifically get rid if there is a last empty column
what do you think is best? or any other idea?

calour/io.py
amnona: fixed

calour/io.py
amnona: fixed

calour/io.py
amnona: i think we should keep none, since this way we can set the default based on the data_file_type while enabling the user to change if needed. if we set default as ',' we won't be able to change with other separators based on file type (say \t if needed)?

calour/io.py
amnona: again, i think maybe we'll need to set row_are_samples to True for openms, so need to be able to set based on data_file_type while still enable user override

calour/io.py
amnona: see above

calour/io.py
amnona: fixed to data_table_sep
also noticed i didn't use it! so fixed (now passed to read() and from there to _read_csv()

amnona:
addressed comments :)
(a few questions remaining)
what do you think?

calour/io.py
RNAer: Reasons:

1. I agree it is needed regularly for MS data thus this parameter and the lines of code should be put in read_ms() instead of read(). I have come across any other usage for this to read other data set.

2. this ID manipulation is very specific to MS data - splitting and getting 1st component. so should be put in read_ms()

RNAer: And if we use it in `read_ms`, rename it to `sample_id_splitter` or `sample_id_sep`?

calour/io.py
```diff
@@ -432,131 +513,105 @@ def read_open_ms(data_file, sample_metadata_file=None, gnps_file=None, feature_m
...
 
     See Also
     --------
     read
     '''
-    logger.debug('Reading OpenMS data (OpenMS bucket table %s, map file %s)' % (data_file, sample_metadata_file))
-    if rows_are_samples:
-        data_file_type = 'openms_transpose'
+    if data_file_type == 'mzmine2':
+        if sample_in_row is None:
```
RNAer: it is not like what you said. for example, in all `data_file_type` branches, when `sample_in_row` is `None`, it is set to `False` in all cases. so you can simplify this lengthy and confusing `if` blocks

calour/io.py
amnona: there are 2 things which i think are getting confused:
1. the "cut_sample_id_sep" variable - while i agree it is useful only in read_ms, i don't see a way we can move it from read. This because otherwise sampleIDs don't match between mapping file and data table, and read() will fail when it tries to link the two (in _read_metadata() line 222 - set_index().
While i agree users should just have their mapping file and table sampleIDs the same, this is not the common practice in MS, and i don't think we should re-educate all MS users (we will just lose them...)

2. the "_sample_id" and "_feature_id" fields which i added to the sample_metadata and feature_metadata respectively. This is not critical and can be removed. However, currently we don't have these values as a column - just the index, and this means we cannot sort/filter/etc. by them. While usually there is another field to sort/filter by, i encountered some cases where i wanted to sort by the sampleID and this is why i added it. Can be removed if you insist but i think it's elegant.

calour/io.py
amnona: agree about sample_in_row, but i need to talk to priya again. i think the default for openms is sample_in_row=True, but not sure. If it is false, will change like you suggest.
regarding other variables, i don't think we have an option but setting them to None. could create a set_none internal function that covers the if to look more elegant. what do you think?

calour/io.py
amnona: fixed. changed cut_sample_id_sep to callable (and also one for feature_id cutting).
we keep the _sample_id and _feature_id fields since they don't bother and may be useful :)

calour/io.py
amnona: fixed to passing a dict (data_table_params) which overrides the (ms file format dependent) defaults.

amnona:
ok, ready for review/merge :)

calour/io.py
```diff
@@ -432,131 +519,95 @@ def read_open_ms(data_file, sample_metadata_file=None, gnps_file=None, feature_m
...
+        ctype = 'biom'
     else:
-        data_file_type = 'openms'
+        raise ValueError("ms data_file_type %s not supported. please use ['mzmine2','openms','gnps-ms2']")
+    param_defaults['use_gnps_id_from_AllFiles'] = True
+    param_defaults['mz_rt_sep'] = '_'
+    # set the default param values if not supplied by the user
+    for cparam, cval in param_defaults.items():
+        data_table_params[cparam] = data_table_params.get(cparam, cval)
+
```
RNAer: this is even more verbose than before. plus, the func param `data_table_params` is ugly.

calour/io.py
```diff
@@ -165,12 +167,19 @@ def _read_open_ms(fp, transpose=True, rows_are_samples=False, sep=','):
...
-    logger.debug('loading OpenMS bucket table %s' % fp)
+    logger.debug('loading csv table %s' % fp)
     # use the python engine as the default (c) engine throws an error
     # a known bug in pandas (see #11166)
     table = pd.read_csv(fp, header=0, engine='python', sep=sep)
+
+    # if the csv file has an additional sep at the end of each line, it cause
+    # pandas to create an empty column at the end. This can cause bugs with the
+    # normalization. so we remove it.
+    if np.all(np.isnan(table[table.columns[-1]])):
```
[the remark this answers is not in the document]
amnona: cool. fixed

calour/filtering.py
```diff
@@ -548,3 +549,25 @@ def filter_ids(exp: Experiment, ids, axis=1, negate=False, inplace=False):
...
+    Parameters
+    ----------
+    axis : 0, 1, 's', or 'f', optional
+        0 or 's' (default) to remove samples where data contains nan, 1 or 'f' to remove features
+    negate : bool, optional
+        negate the filtering
+    inplace : bool, optional
+        False (default) to create a copy of the experiment, True to filter inplace
+    '''
+    nan_pos = np.where(np.isnan(exp.data.sum(axis=1-axis)))[0]
```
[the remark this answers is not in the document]
amnona: removed this function since not using it anymore and don't think that useful

calour/io.py
```diff
@@ -270,6 +281,13 @@ def read(data_file, sample_metadata_file=None, feature_metadata_file=None,
         {'index_col': False}.
     cls : ``class``, optional
         what class object to read the data into (:class:`.Experiment` by default)
+    table_sample_id_proc, table_feature_id_proc: None or callable, optional
```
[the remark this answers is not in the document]
amnona: fixed

calour/io.py
```diff
@@ -393,36 +419,97 @@ def read_amplicon(data_file, sample_metadata_file=None,
...
+    'biom': using a biom table for the metabolite table. featureIDs in the table (first column) can be either MZ_RT (concatenated with a separator),
+    or a unique ID matching the gnps_file ids.
+    'openms': using a csv data table with MZ_RT or unqie ID as featureID (first column). samples can be columns (default) or rows (using the sample_in_row=True parameter)
+    'gnps-ms2': a tsv file exported from gnps, with gnps ids as featureIDs.
 
     Parameters
     ----------
     data_file : str
-        name of the OpenMS bucket table CSV file.
+        The name of the data table (mzmine2 output/bucket table/biom table) containing the per-metabolite frequencies.
```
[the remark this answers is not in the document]
amnona: fixed

calour/io.py
```diff
@@ -393,36 +419,97 @@ def read_amplicon(data_file, sample_metadata_file=None,
...
+    as well as able to read user specified formats.
+    With the installation of the gnps-calour database interface, calour can integrate MS2 information from GNPS into the analysis.
+    If the data table and the gnps file share the same IDs (preferred), GNPS annotations use the uniqueID of the features. Otherwise, calour
+    matches the features to the gnps file using an MZ and RT threshold window (specified by the mz_thresh=XXX, rt_thresh=XXX parameters).
+    Supported formats for ms analysis (as specified by the type='XXX' parameter) include:
+    'mzmine2': using the csv output file of mzmine2. MZ and RT are obtained via the 2nd and 3rd column in the file.
+    'biom': using a biom table for the metabolite table. featureIDs in the table (first column) can be either MZ_RT (concatenated with a separator),
+    or a unique ID matching the gnps_file ids.
+    'openms': using a csv data table with MZ_RT or unqie ID as featureID (first column). samples can be columns (default) or rows (using the sample_in_row=True parameter)
+    'gnps-ms2': a tsv file exported from gnps, with gnps ids as featureIDs.
```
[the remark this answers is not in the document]
amnona: fixed. now looks pretty :)

calour/io.py
```diff
@@ -393,36 +419,97 @@ def read_amplicon(data_file, sample_metadata_file=None,
...
     data_file : str
-        name of the OpenMS bucket table CSV file.
+        The name of the data table (mzmine2 output/bucket table/biom table) containing the per-metabolite frequencies.
     sample_metadata_file : str or None (optional)
         None (default) to not load metadata per sample
-        str to specify name of sample mapping file (tsv)
+        str to specify name of sample mapping file (tsv).
+        Note: sample names in the bucket table and sample_metadata file must match. In case bucket table sample names contains additional
+        information, you can split them at the separator character (usually '_'), keeping only the first part, using the cut_sample_id_sep='_' parameter
+        (see below)
```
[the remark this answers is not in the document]
amnona: nice catch. fixed

calour/io.py
```diff
@@ -393,36 +419,97 @@ def read_amplicon(data_file, sample_metadata_file=None,
...
-        for use with the 'gnps' database in plot
+        name of the gnps clusterinfosummarygroup_attributes_withIDs_arbitraryattributes/XXX.tsv file, for use with the 'gnps' database.
+        This enables identification of the metabolites with known MS2 (for the interactive heatmap and sorting/filtering etc), as well as linking
+        to the gnps page for each metabolite (from the interactive heatmap - by double clicking on the metabolite database information).
+        Note: requires gnps-calour database interface module (see Calour installation instructions for details).
     feature_metadata_file : str or None (optional)
         Name of table containing additional metadata about each feature
         None (default) to not load
+    data_file_type: str, optional
+        the data file format. sets default values for the additional parameters in order to properly load the table (which can be overwritten). options include:
```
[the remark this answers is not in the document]
amnona: i have fixed this problem.

calour/io.py
amnona: fixed.
if you have a better less ugly name - i can update.

calour/io.py
```diff
@@ -393,36 +419,97 @@ def read_amplicon(data_file, sample_metadata_file=None,
...
+            table is a tsv/json/hdf5 biom table, columns are samples.
+        'openms': load an openms output table
+            MZ and RT are obtained via the featureID (first column), which is assumed to be MZ_RT.
+            GNPS linking is indirect via the mz and rt threshold windows.
+            table is a csv table, columns are samples.
+        'gnps-ms2': load a gnps exported biom table
+            MZ and RT are obtained via the gnps_file if available, otherwise are NA
+            GNPS linking is direct via the first column (featureID).
+            table is a tsv/json/hdf5 biom table, columns are samples.
+    data_table_params: dict, optional
```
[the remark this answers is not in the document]
amnona: i don't understand. isn't this what we discussed? we want to have file type dependent default values that can be overwritten. what are you suggesting?

amnona:
I think we should have one function for read_ms, with selecting the file_type for the appropriate table.
i think having a lot of functions can be confusing for users and overload the documents, and also at the end all of them load ms data.
since there isn't a standard for the ms table, even if we create specific read_ms_mzmine() etc. functions, we still need to provide all the parameters to each function (i.e. split sample ids, transpose, gnps_id source etc.) so we won't save a lot of confusion there....

amnona:
ready for review.

amnona:
ok, changed the parameters and default values method for read_ms.
can review/merge (and remember, we know where you live)

Issue #92, opened 2018-04-11
allow save seq collections in jupyter notebook gui interface


tanaes:
it would be really great if selected feature collections could be stored to some sort of variable in the Jupyter interface.

PR #93, opened 2018-04-11
wip: saving features in jupyter notebook

tanaes:
Stub for issue #92

I was able to actually store a new column to the exp dataframe, but don't quite understand the notebook widget interface well enough to store the values correctly. Perhaps [this link](https://stackoverflow.com/questions/26352555/how-does-one-get-widget-values-with-a-button-in-ipython) has some relevant info? 

example_gif

### DIFF
--- a/calour/heatmap/plotgui_jupyter.py
+++ b/calour/heatmap/plotgui_jupyter.py
@@ -10,7 +10,6 @@ from .._doc import ds
 
 logger = getLogger(__name__)
 
-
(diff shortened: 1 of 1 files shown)


tanaes:
@amnona any thoughts on this?

amnona:
waiting to merge the mass spec PR and then will do this :) (should be ready
in a couple of days)
thanks for the suggestion!

PR #97, opened 2018-04-25
fix a bug of colors x/y bars for heatmap

RNAer:
If a dict given to `barx_colors`, the dict was converted to list of
its keys and color info was lost in bar plotting. The problem is that
`_to_list` should convert a dict to a list of dict.

### DIFF
--- a/calour/heatmap/heatmap.py
+++ b/calour/heatmap/heatmap.py
@@ -488,9 +488,9 @@ def plot(exp: Experiment, title=None,
     bary_fields : str or list of str, optional
         column name(s) in sample metadata (barx) / feature metadata (bary). It plots a bar
         for each column. It doesn't plot color bars by default (None)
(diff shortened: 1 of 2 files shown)


calour/heatmap/heatmap.py
```diff
@@ -488,9 +488,9 @@ def plot(exp: Experiment, title=None,
     bary_fields : str or list of str, optional
         column name(s) in sample metadata (barx) / feature metadata (bary). It plots a bar
         for each column. It doesn't plot color bars by default (None)
-    barx_width : float, optional
-    bary_width : float, optional
-        The width of the bars
+    barx_width : float or list of float, optional
+    bary_width : float or list of float, optional
+        The thicknesses of the each bar. The default thickness usually looks good enough.
```
qiyunzhu: Should be `The thickness of each bar.`

PR #98, opened 2018-04-26
add functions for classification

RNAer:


calour/tests/test_training.py
```diff
@@ -43,6 +50,62 @@ def test_split_train_test(self):
...
+        exp = ca.Experiment(X, smd, sparse=False)
+        run = exp.classify('plant', KNeighborsClassifier,
+                           predict='predict_proba',
+                           cv=StratifiedKFold(3, random_state=0))
+        res = next(run)
+        obs = pd.read_table(join(self.test_data_dir, 'iris_result.txt'), index_col=0)
+        pdt.assert_frame_equal(res, obs)
+        # plot_roc(res)
+        # from matplotlib import pyplot as plt
+        # plt.show()
```
amnona: do we want to keep or remove this?

calour/tests/test_training.py
```diff
@@ -43,6 +50,62 @@ def test_split_train_test(self):
...
+        result = pd.read_table(join(self.test_data_dir, 'iris_pred.txt'))
+        ax = plot_roc(result)
+        legend = ax.get_legend()
+        for exp, obs in zip(legend.get_texts(),
+                            ['Luck',
+                             'versicolor (0.95 $\\pm$ 0.06)',
+                             'virginica (0.96 $\\pm$ 0.04)',
+                             'setosa (0.98 $\pm$ 0.01)']):
+            self.assertEqual(exp.get_text(), obs)
+        # from matplotlib import pyplot as plt
```
amnona: again - keep comments or delete?

calour/tests/test_training.py
```diff
@@ -43,6 +50,62 @@ def test_split_train_test(self):
...
+        # plt.show()
+
+    def test_plot_roc_binary(self):
+        result = pd.read_table(join(self.test_data_dir, 'iris_pred.txt'))
+        result['Y_TRUE'] = ['setosa' if i == 'setosa' else 'not setosa'
+                            for i in result['Y_TRUE']]
+        result['not setosa'] = 1 - result['setosa']
+        ax = plot_roc(result, pos_label='setosa')
+        # from matplotlib import pyplot as plt
+        # plt.show()
```
amnona: keep comments?

calour/tests/test_training.py
```diff
@@ -43,6 +50,62 @@ def test_split_train_test(self):
...
+        for exp, obs in zip(legend.get_texts(),
+                            ['Luck',
+                             'setosa (0.98 $\pm$ 0.01)']):
+            self.assertEqual(exp.get_text(), obs)
+
+    def test_plot_cm(self):
+        result = pd.read_table(join(self.test_data_dir, 'iris_pred.txt'))
+        ax = plot_cm(result)
+        from matplotlib import pyplot as plt
+        plt.show()
```
amnona: do we want this plot in the test?

calour/training.py
```diff
@@ -14,11 +14,16 @@
...
 from logging import getLogger
+import itertools
 
 from sklearn.feature_extraction import DictVectorizer
-from sklearn.model_selection import train_test_split
+from sklearn.model_selection import train_test_split, RepeatedStratifiedKFold, ParameterGrid
+from sklearn.model_selection._split import check_cv
+from sklearn.base import is_classifier
+from sklearn.metrics import roc_curve, auc, confusion_matrix
+from matplotlib import pyplot as plt
```
amnona: i think it's a problem to import pyplot here since then pyplot is imported when the calour is loaded, and then it sets the matplotlib backend.
we should import in the function that uses pyplot instead.

calour/training.py
```diff
@@ -121,24 +126,202 @@ def split_train_test(exp: Experiment, field, test_size, train_size=None, stratif
...
+    Parameters
+    ----------
+    field : str
+        column name in the sample metadata, which contains the classes we want to predict.
+    estimator : object
+        scikit-learn estimator. e.g. :class:`sklearn.ensemble.RandomForestClassifer`
+    cv : int, cross-validation generator or an iterable
+        similar to the `cv` parameter in :class:`sklearn.model_selection.GridSearchCV`
+    predict : {'predict', 'predict_proba'}
+        the method to predict the validation sets
```
amnona: can you explain what predict/predict_proba do?

calour/training.py
```diff
@@ -121,24 +126,202 @@ def split_train_test(exp: Experiment, field, test_size, train_size=None, stratif
...
-    '''Do the CV
+def classify(exp: Experiment, field, estimator, cv=RepeatedStratifiedKFold(3, 1),
+             predict='predict_proba', param_grid=None):
+    '''Evaluate classification during cross validation.
+
+    Parameters
+    ----------
+    field : str
+        column name in the sample metadata, which contains the classes we want to predict.
+    estimator : object
```
amnona: maybe set sklearn.ensemble.RandomForestClassifer as the default value for the estimator? This way, people who don't know a lot about machine learning can use the classify function without needing to understand all the classifier choices (random forest is a good default for microbiome....)?

calour/training.py
```diff
@@ -121,24 +126,202 @@ def split_train_test(exp: Experiment, field, test_size, train_size=None, stratif
...
+            pred = getattr(model, predict)(X[test])
+            if pred.ndim > 1:
+                df = pd.DataFrame(pred, columns=model.classes_)
+            else:
+                df = pd.DataFrame(pred, columns=['Y_PRED'])
+            df['Y_TRUE'] = y[test].values
+            df['SAMPLE'] = y[test].index.values
+            df['CV'] = i
+            dfs.append(df)
+        yield pd.concat(dfs, axis=0).reset_index(drop=True)
```
amnona: why does it do yield? what is the meaning to do next? another cross validation? maybe worth explaining in the doc? (i'm confused....)

calour/training.py
```diff
@@ -121,24 +126,202 @@ def split_train_test(exp: Experiment, field, test_size, train_size=None, stratif
...
+    ----------
+    field : str
+        column name in the sample metadata, which contains the classes we want to predict.
+    estimator : object
+        scikit-learn estimator. e.g. :class:`sklearn.ensemble.RandomForestClassifer`
+    cv : int, cross-validation generator or an iterable
+        similar to the `cv` parameter in :class:`sklearn.model_selection.GridSearchCV`
+    predict : {'predict', 'predict_proba'}
+        the method to predict the validation sets
+    param_grid : dict of string to sequence, or sequence of such
```
amnona: i don't understand what "dict of string to sequence" means... :(

calour/training.py
amnona: also, what happens if param_grid is None (default)? do we get any results?
In general, naively i would expect the classify function to return and not yield... but i am not an expert random forester :)

calour/training.py
```diff
@@ -121,24 +126,202 @@ def split_train_test(exp: Experiment, field, test_size, train_size=None, stratif
...
+            dfs.append(df)
+        yield pd.concat(dfs, axis=0).reset_index(drop=True)
+
+
+def plot_cm(result, normalize=False, title='confusion matrix', cmap=plt.cm.Blues, ax=None):
+    '''Plot confusion matrix
+
+    Parameters
+    ----------
+    result : pandas.DataFrame
```
amnona: maybe change parameter name to classify_df or something else?
result as an input is a bit confusing...

calour/training.py
```diff
@@ -121,24 +126,202 @@ def split_train_test(exp: Experiment, field, test_size, train_size=None, stratif
...
+        y_pred = result['Y_PRED']
+    return confusion_matrix(result['Y_TRUE'], y_pred, labels=labels, **kwargs)
+
+
+def plot_roc(result, pos_label=None, title='ROC', cmap=plt.cm.Dark2, ax=None):
+    '''Plot ROC curve.
+
+    Parameters
+    ----------
+    result : pandas.DataFrame
```
amnona: again maybe change name?

calour/training.py
```diff
@@ -121,24 +126,202 @@ def split_train_test(exp: Experiment, field, test_size, train_size=None, stratif
...
+        data frame containing predictions per sample (in row). It must have a column of
+        true class named "Y_TRUE" and multiple columns of predicted probabilities for each class.
+    title : str
```
amnona: what about pos_label?

calour/training.py
```diff
@@ -121,24 +126,202 @@ def split_train_test(exp: Experiment, field, test_size, train_size=None, stratif
...
+    -------
+    matplotlib.axes.Axes
+        The axes for the ROC
+
+    '''
+    if ax is None:
+        fig, ax = plt.subplots()
+
+    ax.axis('equal')
+    ax.plot([0, 1], [0, 1], linestyle='-', lw=1, color='black', label='Luck', alpha=.5)
```
amnona: "Luck" or "Random"?

amnona:
cool!
a few comments.

calour/training.py
RNAer: I remember this - but why its setting backend is a problem? as long as it doesn't change the backend, it's fine?

calour/training.py
RNAer: nope. then they should read the docstring and probably shouldn't use it because this is a analysis requiring careful understanding

calour/training.py
RNAer: added more doc in param_grid and docstring

calour/training.py
RNAer: i think `string to sequence` means the dict map of a string to a sequence. copied from sklearn doc. I prefer not changing it.

calour/training.py
RNAer: i don't think classify_df is better. updated docstring, which should make it clearer

calour/training.py
RNAer: "Luck"

calour/training.py
amnona: i think the problem is with the qt5 gui vs. the notebook gui... don't remember exactly...
anyway, best to move it into the functions that need it :)

PR #101, opened 2018-04-29
Bugfixes

amnona:
minor bugfixes


calour/analysis.py
```diff
@@ -160,9 +165,18 @@ def diff_abundance(exp: Experiment, field, val1, val2=None, method='meandiff', t
     Returns
     -------
     Experiment
-        A new experiment with only significant (FDR <= maxfval) difference, sorted according to difference
+        A new experiment with only significant (FDR <= maxfval) difference, sorted according to the effect size.
 
+        The new experiment contains additional feature_metadata_fields that include:
```
RNAer: put an empty line here

calour/analysis.py
```diff
@@ -160,9 +165,18 @@ def diff_abundance(exp: Experiment, field, val1, val2=None, method='meandiff', t
...
+            '_calour_diff_abundance_pval' : the p-value for the feature
```
RNAer: make these a list and delete empty lines

calour/filtering.py
```diff
@@ -74,8 +74,14 @@ def downsample(exp: Experiment, field, axis=0, num_keep=None, inplace=False):
     logger.debug('downsample on field %s' % field)
     if axis == 0:
         x = exp.sample_metadata
+        error_axis_name = 'sample'
     elif axis == 1:
         x = exp.feature_metadata
+        error_axis_name = 'feature'
+
+    if field not in x:
```
RNAer: why adding this? what's wrong with just let error happen? it decreases test coverage and provides no extra benefit?

amnona: i like it better this way :)
1. the error message is cryptic for users who don't know the code... it gives something from pandas and not clear what the problem is.
2. Also for me, the new error message lets me see the real metadata field names, so i can see my problem without printing the dataframe etc.
3. Can extend the coverage by testing the error message raised, but i don't think critical.

calour/analysis.py
amnona: fixed

calour/analysis.py
amnona: is this what you meant?

amnona:
ok, addressed comments.
(still like the new error msg better - see reasons in PR)

calour/analysis.py
```diff
@@ -168,11 +168,12 @@ def diff_abundance(exp: Experiment, field, val1, val2=None, method='meandiff', t
...
         The new experiment contains additional feature_metadata_fields that include:
-            '_calour_diff_abundance_pval' : the p-value for the feature
-
-            '_calour_diff_abundance_effect' : the effect size (t-statistic)
 
+            [
+            '_calour_diff_abundance_pval' : the p-value for the feature,
+            '_calour_diff_abundance_effect' : the effect size (t-statistic),
             '__calour_diff_abundance_group' : the value (in field) where the statistic is higher
+            ]
```
RNAer: noooooooo. not python list. like this:

```
- a
- b
- c
```

calour/filtering.py
RNAer: alright. i am not convinced, but fine to keep it.

amnona: :)

PR #103, opened 2018-04-30
Export html

amnona:
add the exp.export_html() command.
it creates an interactive html heatmap.
and it comes with unit tests :)


calour/tests/test_export_html.py
```diff
@@ -0,0 +1,42 @@
...
+        self.test1 = ca.read(self.test1_biom, self.test1_samp, normalize=None)
+
+    def test_export_html(self):
+        exp = self.test1
+        d = mkdtemp()
+        f = join(d, 'calour_interactive_heatmap.html')
+        exp.export_html(sample_field='group', output_file=f, feature_field='taxonomy', title='the heatmap')
+        # load the results
+        with open(f) as output_fl:
+            output_data = output_fl.readlines()
```
RNAer: https://docs.python.org/3.6/library/tempfile.html#tempfile.NamedTemporaryFile is better than creating temp dir and then a file inside it and deleting it at the end.

calour/tests/test_export_html.py
```diff
@@ -0,0 +1,42 @@
...
+        with open(join(self.test_data_dir, 'export_html_result.html')) as expected_fl:
+            expected_data = expected_fl.readlines()
+        # we cannot test equality of the whole files since plot output is different
+        # on different machines
```
RNAer: why it is not deterministic? any randomization in the process?

RNAer:
could you explain briefly what the changes in template do?

calour/tests/test_export_html.py
amnona: It's deterministic, but we store the heatmap as png and embed it into the html. and i think the png is different between different os/computer configs (i.e. bit depth, exact colors, etc.)
when i compared the complete html, it works on my local machine but not on travis, and the problem is in the embedded image...
so no we just test the first 100 lines which contain sample ids etc.

amnona:
addressed the comments (but didn't change anything... see text there)
regarding the template,
mostly changed to add shift+, shift_ to zoom in on sample axis and arrows for moving on sample axis as well. This because in the new chrome version, they disabled the shift+scroll wheel in d3.zoom object, so need to enable another way to zoom in.

calour/tests/test_export_html.py
RNAer: alright. didn't notice you open it in the export_html function too.

PR #104, opened 2018-05-04
refactor filter abundance functions

RNAer:
also add deprecated decorator

### DIFF
--- a/calour/filtering.py
+++ b/calour/filtering.py
@@ -470,7 +470,7 @@ def filter_abundance(exp: Experiment, cutoff=0.01, field=None, **kwargs):
 
     '''
     if exp.normalized <= 0:
-        logger.warning('Do you forget to normalize your data? It is required to run this function')
+        logger.warning('Do you forget to normalize your data? It is required before running this function')
 
     cutoff = exp.normalized * cutoff
 
@@ -479,10 +479,11 @@ def filter_abundance(exp: Experiment, cutoff=0.01, field=None, **kwargs):
     else:
         select = np.zeros(exp.shape[1], dtype='?')
         # convert it to unicode str in case there are NaNs
-        samples = exp.sample_metadata[field].values.astype('U')
-        groups, indices = np.unique(samples, return_inverse=True)
-        for grp in groups:
-            select = select | _mean_abundance(exp.data[indices == grp], axis=0, cutoff=cutoff)
+        values = exp.sample_metadata[field].values.astype('U')
+        groups, indices = np.unique(values, return_inverse=True)
+        for i, _ in enumerate(groups):
+            # Here axis is 0 not 1.
+            select = select | _mean_abundance(exp.data[indices == i], axis=0, cutoff=cutoff)
         return exp.reorder(select, axis=1, **kwargs)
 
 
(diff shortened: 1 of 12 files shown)


calour/filtering.py
```diff
@@ -207,17 +213,18 @@ def filter_by_data(exp: Experiment, predicate, axis=0, negate=False, inplace=Fal
...
+        * 'freq_ratio': calls :func:`.freq_ratio`,
+        * 'unique_cut': calls :func:`.unique_cut`
     axis : 0, 1, 's', or 'f', optional
-        Apply predicate on each row (samples) (0) or each column (features) (1)
+        Apply predicate on each row (ie samples) (0, 's') or each column (ie features) (1, 'f')
+    field : str or `None`, optional
+        The column in the sample_metadata (feature_metadata). If it is not `None`, the
+        features that has mean abundance lower than the cutoff in *ALL*
+        sample categories in the specified sample_metadata column will
+        be filtered away.
```
amnona: In "features that has mean abundance lower than the cutoff...", is the field for mean_abundance or for any predicate?
Also should be "that have" rather than "that has"

calour/filtering.py
```diff
@@ -430,51 +465,64 @@ def filter_samples(exp: Experiment, field, values, negate=False, inplace=False):
...
+    return exp.filter_by_data('mean_abundance', axis=1, field=field, cutoff=cutoff, **kwargs)
 
 
 @ds.with_indent(4)
 @Experiment._record_sig
-def filter_prevalence(exp: Experiment, fraction, cutoff=1/10000, **kwargs):
-    '''Filter features keeping only ones present in at least fraction fraction of the samples.
-    This is a convenience function wrapping filter_by_data()
+def filter_abundance(exp: Experiment, cutoff=10, **kwargs):
+    '''Filter features with a mean at least cutoff of the mean total abundance/sample
```
amnona: I think this is a very confusing explanation (mean of the mean total abundance/sample). I couldn't understand it and i'm supposed to know what's going on :)
Maybe something like:
Filter features with sum over all samples at least cutoff.

calour/filtering.py
```diff
@@ -430,51 +465,64 @@ def filter_samples(exp: Experiment, field, values, negate=False, inplace=False):
...
+
+    For example, to keep features with mean abundance of 1% use `filter_abundance(cutoff=0.01)`.
 
     Parameters
     ----------
-    fraction : float
-        Keep features present at least in fraction of samples
     cutoff : float, optional
-        The minimal fraction of reads for the otu to be called present in a sample
+        The minimal mean abundance fraction (out of the mean of total
```
amnona: again, very confusing to me

calour/tests/test_amplicon_experiment.py
```diff
@@ -39,6 +39,7 @@ def test_filter_taxonomy(self):
 
     def test_filter_fasta(self):
         # test keeping the sequences from fasta
+        print(self.seqs1_fasta)
```
amnona: delete this print?

amnona:
few small comments
:)

RNAer:
thanks for pointing out the outdated doc. corrected. it's ready

PR #107, opened 2018-05-11
refactor downsample

RNAer:


calour/filtering.py
```diff
@@ -82,37 +82,44 @@ def downsample(exp: Experiment, field, axis=0, num_keep=None, inplace=False):
...
+    Parameters
+    ----------
+    x : array
+    n : int. count
+
+    Returns
+    -------
+    array of bool
+    '''
+    rand = np.random.RandomState(random_state)
```
amnona: if ranom_state is None, do we want to use RandomState(None) which uses the current time etc. or do we want to skip RandomState() all together?
(in the latter case, if a user does an initial random seed, all following results will be reproducible, whereas if we have RandomState(None), it will vary.
I like more to do the RandomState() call only if random_state is not None.

amnona:
cool. one small comment :)

PR #109, opened 2018-05-16
add new plottings

RNAer:
tried to add global random seed but decided to revert it.


calour/__init__.py
```diff
@@ -25,6 +25,9 @@
            'Experiment', 'AmpliconExperiment', 'MS1Experiment',
            'set_log_level']
 
+# this is the global random seed; if it's None, it is unset
+SEED = None
```
amnona: i think we need to get rid of this?

calour/filtering.py
```diff
@@ -42,7 +42,7 @@
 
 from . import Experiment
 from ._doc import ds
-from .util import _to_list
+from .util import _to_list, _set_random_state
```
amnona: again, need to get rid of _set_random_state

calour/plotting.py
```diff
@@ -530,12 +538,25 @@ def plot_scatter_matrix(exp: Experiment, field, feature_ids, title_field=None,
         the height of each figure panel in inches
     aspect : numeric
         Aspect ratio of each figure panel, so that aspect * size gives its width
+
```
amnona: what about doc for plot=None parameter

calour/plotting.py
```diff
@@ -530,12 +538,25 @@ def plot_scatter_matrix(exp: Experiment, field, feature_ids, title_field=None,
...
+    matplotlib.figure.Figure
     '''
     from matplotlib import pyplot as plt
 
     x = exp.sample_metadata[field].values
-    if transform_x is not None:
-        x = transform_x(x)
+
+    if plot is None:
+        if True:
```
amnona: why if True?

calour/util.py
```diff
@@ -504,3 +504,13 @@ def deprecated_func(*args, **kwargs):
             return func(*args, **kwargs)
         return deprecated_func
     return deprecated_decorator
+
+
+def _set_random_state(rand):
```
amnona: delete this function?

amnona:
a few small comments :)

calour/plotting.py
```diff
@@ -517,25 +525,36 @@ def plot_scatter_matrix(exp: Experiment, field, feature_ids, title_field=None,
 
     Parameters
     ----------
-    field : str
+    fields : str
```
amnona: I think it should be 'field' and not 'fields'? since it contains the name of only one column (i.e. usually i'd expect fields to be a list of column names)?

amnona:
one super small comment.
do you want me to "create a merge commit" instead of "squash and merge"

RNAer:
yes, pls create a merge commit

calour/plotting.py
RNAer: nope. plan to make it do multi-task prediction (predict multi labels together). will finish it in next pr. this change doesn't impact current function.

PR #110, opened 2018-05-18
Ms fix

amnona:
fix _split_sample_ids for read_ms when the separator is ' ' (space) not to split the 'row m/z' and 'row retention time' (from mzmine2)
and small typo fix


calour/io.py
```diff
@@ -753,7 +753,7 @@ def _create_biom_table_from_exp(exp, add_metadata='taxonomy', to_list=False):
     return table
 
 
-def _split_sample_ids(sid, split_char=None):
+def _split_sample_ids(sid, split_char=None, ignore_split=['row m/z', 'row retention time']):
```
RNAer: don't put mutable as default value.

calour/manipulation.py
```diff
@@ -224,8 +224,8 @@ def join_experiments(exp: Experiment, other, field_name='experiments', prefixes=
     # when both experiments contain the same sample ids)
     if len(exp.sample_metadata.index.intersection(other.sample_metadata.index)) > 0:
         if prefixes is None:
-            raise ValueError('You need provide prefix to add to sample IDs '
-                             'because the two experiments has some same sample IDs.')
+            raise ValueError('You need provide prefixes=() to add to sample IDs '
```
RNAer: confusing. reading this error msg, I would literally put `prefix=()` in the function call. just use \`prefix\`

amnona:
addressed comments

PR #119, opened 2018-07-14
Fix centroid

amnona:
fix bug for transform=None and sparse experiment
and tidy up code


calour/sorting.py
```diff
@@ -82,13 +82,14 @@ def sort_centroid(exp: Experiment, transform=log_n, inplace=False, **kwargs):
...
     logger.debug('sorting features by center of mass')
     if transform is None:
-        data = exp.data
+        newexp = exp
     else:
         logger.debug('transforming data using %r' % transform)
         newexp = deepcopy(exp)
-        data = transform(newexp, **kwargs).data
+        newexp = transform(newexp, **kwargs)
+    data = newexp.get_data(sparse=False)
```
RNAer: why need to be dense?

amnona: if it's not dense, line 94 :
center_mass = np.divide(center_mass, data.sum(axis=1))

returns a matrix instead of an array since data.sum(axis=1) returns a matrix for sparse data
we can either unpsarse it like in the PR, or we can do:
data_sum = data.sum(axis=1)
if newexp.sparse:
    # convert the matrix to array
    data_sum = data_sum.A1
center_mass = np.divide(center_mass, data_sum)

what do you like better?

RNAer: this is better because you don't need to densify the whole array?

RNAer:
close in favor of #127

PR #127, opened 2018-08-01
improve sort_centroid

RNAer:


calour/io.py
```diff
@@ -402,12 +402,8 @@ def read_amplicon(data_file, sample_metadata_file=None,
     exp = read(data_file, sample_metadata_file, cls=AmpliconExperiment,
                normalize=None, **kwargs)
 
-    # exp.feature_metadata.index = exp.feature_metadata.index.str.upper()
-
     if 'taxonomy' in exp.feature_metadata.columns:
         exp.feature_metadata['taxonomy'] = _get_taxonomy_string(exp, remove_underscore=False)
-    else:
-        exp.feature_metadata['taxonomy'] = 'NA'
```
amnona: I think we need the 'NA' for taxonomy since in the plot for amplicon experiment, we show the taxonomy field, and if the biom table did not contain taxonomy, it will result in error...

calour/transforming.py
```diff
@@ -317,7 +317,7 @@ def random_permute_data(exp: Experiment, normalize=True):
 
 
 @Experiment._record_sig
-def center_log(exp: Experiment, method=lambda matrix: matrix + 1, centralize=False, inplace=False):
+def center_log_ratio(exp: Experiment, method=lambda matrix: matrix + 1, centralize=False, inplace=False):
```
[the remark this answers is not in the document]
RNAer: updated

calour/io.py
RNAer: no, I looked - we don't need it for default options. we shouldn't have a useless NA column anyway.

PR #130, opened 2018-08-10
add a notebook for machine learning

RNAer:


amnona:
cool.
can you also add an example for ROC curve/comparison to random model (maybe RF classifier rather than regression...)?

PR #131, opened 2018-08-13
add classification example in the notebook

RNAer:


amnona:
cool. nice example.
two small things:
1. the testing is failing because of the changed  missing sample message:
Traceback (most recent call last):
  File "/home/travis/build/biocore/calour/calour/tests/test_io.py", line 119, in test_read_feature_kwargs
    self.assertRegex(cm.output[1], "dropped: {'SAMPLE_NOT_EXIST'}")
AssertionError: Regex didn't match: "dropped: {'SAMPLE_NOT_EXIST'}" not found in "WARNING:calour.io:These have metadata but do not have data - dropped (1): {'SAMPLE_NOT_EXIST'}"

2. in the notebook, in the classification part (after 34), there is text which is not related:
"We can plot out the result as following. Each dot is a sample with observed and predicted pH, colored by the fold of cross validation the sample is from. The diagonal line indicates perfect predition. The correlation coefficient and its p-value between the prediction and observation are also annotated around the top of the plot."

should be removed :)

PR #133, opened 2018-10-05
add contribution guidelines

RNAer:


RNAer:
let's fix it. i am not sure what's the problem either. let's seek solution
both.

PR #134, opened 2018-10-07
Improve dbbact terms and add common error msgs

amnona:
- Add specific error messages about common metadata loading problems (i.e. duplicate ids or other metadata read problem).
- add database specific kwargs to the add_terms_to_features()
- make the add_terms_to_features work with a dict of {term:value} returned by the database interface. rather than a list of terms which is counted (more general and efficient).


calour/io.py
```diff
@@ -235,6 +239,13 @@ def _read_metadata(ids, f, kwargs):
             logger.warning('These have data but do not have metadata: %r' % diff)
         # reorder the id in metadata to align with biom
         # metadata = metadata.loc[ids, ]
+
+        # check if we have duplicate index ids (which will raise)
+        dup = metadata.index.duplicated(keep=False)
+        if dup.sum() > 0:
+            errmsg = '%d duplicate id values encountered in column %s of mapping file %r:' % (dup.sum(), metadata.index.name, f)
```
RNAer: metadata.index.name can be empty. it's better to say "1st column" or "index column"

calour/io.py
```diff
@@ -224,7 +224,11 @@ def _read_metadata(ids, f, kwargs):
         else:
             kwargs['dtype'] = {index_col: str}
 
-        metadata = pd.read_table(f, **kwargs)
+        try:
+            metadata = pd.read_table(f, **kwargs)
```
RNAer: why this instead of letting it fail?

calour/database.py
```diff
@@ -363,8 +371,9 @@ def get_feature_terms(self, features, exp=None):
 
         Returns
         -------
-        feature_terms : dict of list of str/int
-            key is the feature, list contains all terms associated with the feature
+        feature_terms : dict of {feature(str): dict term_scores {term(str): score(float)}}
```
RNAer: this is hard to understand

calour/database.py
```diff
@@ -99,6 +98,7 @@ def add_terms_to_features(exp: Experiment, dbname, use_term_list=None, field_nam
         None to get default type
     ignore_exp : list of int or None, optional
         list of experiments to ignore when adding the terms
+    **kwargs: database specific additional parameters (see database interface get_feature_terms() function for specific terms)
```
RNAer: could you provide a reference to link the function (using the rst syntax as did in doc in other functions)?

calour/database.py
amnona: it's duplicated in the second PR (export_html) - fixed it there... sorry

calour/database.py
amnona: I think it's more problematic since it depends on the database modules you have installed. So the kwargs are different between dbbact_calour and gnps_calour. Also, these modules reside in different github repos, so don't know how to do it...

calour/io.py
amnona: changed to take care of both cases :)

calour/io.py
amnona: I saw several users encountering a problem with reading the metadata, and the regular error message is confusing for non-advanced python users (who don't know the calour code). This just gives them a general message the mapping file has a problem and then raises the regular python error, so less confusing for them (the error message is the first line they see when it fails...).

amnona:
@RNAer ready for re-review :)

calour/io.py
```diff
@@ -243,7 +243,11 @@ def _read_metadata(ids, f, kwargs):
         # check if we have duplicate index ids (which will raise)
         dup = metadata.index.duplicated(keep=False)
         if dup.sum() > 0:
-            errmsg = '%d duplicate id values encountered in column %s of mapping file %r:' % (dup.sum(), metadata.index.name, f)
+            column_str = '%d' % index_col
+            if metadata.index.name is not None:
+                if len(metadata.index.name) > 0:
+                    column_str = metadata.index.name
+            errmsg = '%d duplicate id values encountered in index column %s of mapping file %r:' % (dup.sum(), column_str, f)
```
RNAer: why making it so complicated?  only need to say the index column has duplicate IDs? you don't really need the name of index column, right?

amnona: i think users will find it easier if they get the column name (if it exists) rather than needing to count to column number XXX.
the complication is in the code (and it's not THAT complicated...)
but if you feel strongly to just give the column number instead, can also reduce to that (but the users will be sad....)

amnona:
@RNAer  try again?

PR #135, opened 2018-10-08
Html export fix

amnona:
change the data transform to imshow normalization similar to heatmap()
also, the transform was not done on a copy of the data in some cases, so was also a bug in the old transform implementation


calour/database.py
```diff
@@ -363,8 +363,9 @@ def get_feature_terms(self, features, exp=None):
 
         Returns
         -------
-        feature_terms : dict of list of str/int
-            key is the feature, list contains all terms associated with the feature
+        feature_terms : dict of {feature(str): dict term_scores {term(str): score(float)}}
```
RNAer: this line is hard to understand

amnona:
@RNAer : fixed the docstring. Is this better?

PR #138, opened 2018-11-24
improve documentation

RNAer:


calour/training.py
```diff
@@ -692,7 +692,26 @@ def plot_roc(result, classes=None, title='ROC', cv=True, cmap=None, ax=None):
     ax.set_title(title)
     ax.legend(loc="lower right")
 
-    return ax
+    return ax, mean_auc if cv else roc_auc
+
+
+def plot_calibration(y_true, y_prob, bins=10):
```
amnona: What does this function do? does it have a docstring? or is it internal (maybe should start with "_")?
Do we want also to add it to the sphinx autodoc?

amnona:
one small comment :)

PR #139, opened 2018-12-13
Experiment optional mutable

amnona:
Had the wonderful mutable as optional param bug in the ca.Experiment __init__ :)
Also fixed the feature_metadata in this function to contain an index if no feature_metadata is supplied (if feature_metadata is empty or None, other calour functions don't work).


RNAer:
told you not to use mutable as default arg...

PR #140, opened 2019-01-02
Improve term plot

amnona:
improve the term plot to include number of enriched experiments


amnona:
@RNAer can you merge?

calour/filtering.py
```diff
@@ -458,7 +458,10 @@ def filter_samples(exp: Experiment, field, values, negate=False, inplace=False):
         the filtered object
 
     '''
-    values = _to_list(values)
+    # if it is None - pass to filter_by_metadata so will remove the NaN samples
```
RNAer: i think this behavior is not intuitive.

calour/plotting.py
```diff
@@ -93,6 +93,9 @@ def plot_enrichment(exp: Experiment, enriched, max_show=10, max_len=40, ax=None,
         name for terms enriched in group1 or group2 respectively, or None to not show legend
     colors: tuple of (str, str) or None (optional)
         Colors for terms enriched in group1 or group2 respectively
+    enriched_exp_color: str or None, optional
+        If not None, the color to show the number of enriched experiments for each term in the bar.
+        None to not show the enriched experiments count
```
RNAer: what does the default 'w' mean?

ci/conda_requirements.txt
```diff
@@ -5,7 +5,7 @@ numpy
 scipy
 pyqt>=5
 pandas
-matplotlib
+matplotlib=2.1.2
```
RNAer: remove this

calour/plotting.py
```diff
@@ -164,6 +183,9 @@ def plot_diff_abundance_enrichment(exp: Experiment, max_show=10, max_len=40, ax=
         True to show the color legend, False to hide it
     **kwargs : dict, optional
         Additional database specific enrichment parameters (see per-database module documentation for .enrichment() method)
+    enriched_exp_color: str or None, optional
```
RNAer: same with 'w'

calour/filtering.py
amnona: We need a convenient way to filter samples with None in metadata (since read() also reads samples from the biom table not present in the mapping file and puts None.
We have this behavior for filter_by_metadata() we just pass the parameter.
i think it is good. any other suggestions?

calour/plotting.py
amnona: white. the background color is the bar color (red/green) so white looks nice.
added explanation and changed 'w' to 'white' in PR

ci/conda_requirements.txt
amnona: same is prev. comment

calour/plotting.py
amnona: same as prev. color comment. fixed

amnona:
addressed the comments

RNAer:
I'll ask for before vs. after snapshots for plotting changes next time.

PR #150, opened 2019-04-10
change the output columns of correlation() and diff_abundance()

RNAer:


calour/analysis.py
```diff
@@ -301,8 +323,8 @@ def _new_experiment_from_pvals(cexp, exp, keep, odif, pvals):
     Returns
     -------
     Experiment
-    Containing only significant features, sorted by effect size.
-    Each feature contains 2 new metadata fields: _calour_diff_abundance_pval, _calour_diff_abundance_effect
+        Containing only significant features, sorted by effect size.
+        Each feature contains 2 new metadata fields: "_calour_pval", "_calour_stat"
```
amnona: hard coded in the doc? or maybe specify names defined by _CALOUR_STAT etc.

amnona:
one small comment...
do you still like the global field names even though it complicates the docstring? i don't have a strong opinion...

amnona:
also need to fix plotting.py to use the new names using the global variables - can you fix it?

RNAer:
updated.

PR #152, opened 2019-05-07
Fix history

amnona:
Fix a bug when we have 3 nested calls history also records the 3rd
for example when using exp.diff_abundance(), we get in the history filter_ids() coming from within the diff_abundance nesting....


RNAer:
pls show the bug with code and output in an example.

amnona:
code:
import calour as ca

test1 = ca.read('../calour/tests/data/test1.biom', '../calour/tests/data/test1.sample', normalize=None)
test1 = test1.cluster_features(10)
dd=test1.diff_abundance('group','1','2')
for idx,x in enumerate(dd._call_history):
    print('%d : %s' % (idx, x))

old (bug) version output:
0 : filter_by_data('abundance', 1, None, cutoff=10, inplace=False, mean_or_sum='sum')
1 : filter_abundance(10, inplace=False)
2 : cluster_data(<function transform at 0x1a183fc158>, 1, inplace=False, scale__axis=1, steps=[<function log_n at 0x1a183f0f28>, <function scale at 0x1a183f0d08>])
3 : cluster_features(10)
4 : filter_ids(array(['AC', 'AT', 'AA', 'TA', 'TC', 'AG', 'GG'], dtype=object))
5 : diff_abundance('group', '1', '2')

new (fixed) version output:
0 : cluster_features(10)
1 : diff_abundance('group', '1', '2')

NOTE:  in old version, the filter_abundance in step 1 and cluster_data in step 2 are nested from cluster_features(10), and the filter_ids in step 5 is nested in the diff_abundace.

PR #154, opened 2019-05-28
Load qiime2

amnona:
Added read_qiime2() function that can also use the rep_seqs.qza and taxonomy.qza qiime2 artifacts.
Change the read function to use unzip instead of the qiime2 module. this enables running the read_qiime2 even when qiime2 environment is not active
Added unit tests (the old read_qiime2 unit test did not work in travis due to the need to install qiime2)


calour/_testing.py
```diff
@@ -54,8 +54,13 @@ def setUp(self):
         self.gnps_cluster_info = join(test_data_dir, 'gnps_clusterinfosummarygroup.txt')
         # a fasta file for testing the AmpliconExperiment
         self.seqs1_fasta = join(test_data_dir, 'seqs1.fasta')
-        # a qiime2 biom table artifact
+        # a qiime2 non-hashed biom table artifact
```
RNAer: do we really need a new data set? how about using one of the existing?

calour/io.py
```diff
@@ -80,13 +84,50 @@ def _read_biom(fp, transpose=True):
...
+    internale_data: str
+        the internal qiime2 qza file name (i.e. 'data/feature-table.biom' for biom table etc.)
+
+    Returns
+    -------
+    str: name of data file inside the tempdir
+    '''
+    if not zipfile.is_zipfile(data_file):
+        raise ValueError('the qiime2 file %s is not a valid zip file. Is it a qiime2 artifact (.qza) file?' % data_file)
+    fl = zipfile.ZipFile(data_file)
```
RNAer: this should use `with`, like open a file object

calour/io.py
```diff
@@ -80,13 +84,50 @@ def _read_biom(fp, transpose=True):
...
+    internal_name = None
+    for fname in fl.namelist():
+        if fname.endswith(internal_data):
```
RNAer: this should be `==` instead of `endswith`?

RNAer: btw, you should use the following?

````
with ZipFile('spam.zip') as myzip:
    with myzip.open('eggs.txt') as myfile:
        print(myfile.read())
````

File: calour/io.py [hunk 4]
```diff
@@ -80,13 +84,50 @@ def _read_biom(fp, transpose=True):
...
 
     Parameters
     ----------
     fp : str
         file path to the qiime2 (.qza) biom table
+    rep_seq_file: str or None, optional
+        If not none, name of the representative sequencees qiime2 artifact, used to translate sequnece hashes
+        to the actual sequences.
+    taxonomy_file: str or None, optional
+        If not none, name of the taxonomy qiime2 artifact file.
```
RNAer: these parameters are not in the function API

calour/io.py
```diff
@@ -103,23 +144,58 @@ def _read_qiime2(fp, transpose=True):
...
-    data = table.matrix_data
-    feature_md = _get_md_from_biom(table)
+    return sid, fid, data, fmd
 
-    if transpose:
-        logger.debug('transposing table')
-        data = data.transpose()
 
-    return sid, fid, data, feature_md
+def read_qiime2(fp, sample_metadata_file=None, rep_seq_file=None, taxonomy_file=None, use_qiime2=False, **kwargs):
```
RNAer: no need for `use_qiime2`?

calour/io.py
```diff
@@ -103,23 +144,58 @@ def _read_qiime2(fp, transpose=True):
...
+        if not None, file path to the qiime2 representative sequences artifact file (defined by the qiime2 --o-representative-sequences parameter)
+    taxonomy_file: None or str, optional
+        if not None, add taxonomy for each feature using the qiime2 taxonomy artifact file (output of the qiime2 feature-classifier command)
+
+    Keyword Arguments
+    -----------------
+    %(io.read.parameters)s
+    '''
+    newexp = read_amplicon(fp, sample_metadata_file=sample_metadata_file, data_file_type='qiime2', **kwargs)
+    with tempfile.TemporaryDirectory() as tempdir:
```
RNAer: if using zipfile.open, we don't need temp dir?

calour/io.py
```diff
@@ -103,23 +144,58 @@ def _read_qiime2(fp, transpose=True):
...
+        # if rep-seqs file is supplied, translate hashes to sequences
+        if rep_seq_file is not None:
+            logger.debug('loading rep_seqs file %s' % rep_seq_file)
+            rep_seqs = []
+            rs_name = _file_from_zip(tempdir, rep_seq_file, internal_data='data/dna-sequences.fasta')
+            for cseq in skbio.read(rs_name, format='fasta'):
+                cseq = str(cseq).upper()
```
RNAer: do we really want `upper`? sometimes cases indicates something meaningful

calour/io.py
```diff
@@ -103,23 +144,58 @@ def _read_qiime2(fp, transpose=True):
...
+            newexp.feature_metadata['_feature_id'] = rep_seqs
+            newexp.feature_metadata.set_index('_feature_id', inplace=True)
+
+        # if taxonomy file is supplied, load it into the feature metadata
+        if taxonomy_file is not None:
+            logger.debug('loading taxonomy file %s' % taxonomy_file)
+            tax_name = _file_from_zip(tempdir, taxonomy_file, internal_data='data/taxonomy.tsv')
+            taxonomy_df = pd.read_table(tax_name)
+            taxonomy_df.set_index('Feature ID', inplace=True)
+            newexp.feature_metadata = pd.concat([newexp.feature_metadata, taxonomy_df], axis=1)
```
RNAer: does concat align row index? in case it is not matched.

calour/tests/test_io.py
```diff
@@ -230,11 +230,18 @@ def test_read_open_ms_samples_rows(self):
...
+        self.assertEqual(exp.shape, (104, 658))
+
+        # and the hashed table with rep seqs and taxonomy files
+        exp = ca.read_qiime2(self.q2_cfs_table, sample_metadata_file=self.q2_cfs_map,
+                             rep_seq_file=self.q2_cfs_repseqs, taxonomy_file=self.q2_cfs_taxonomy,
+                             normalize=None, min_reads=None)
+        self.assertEqual(exp.shape, (87, 2130))
+        # test if the index is indeed sequences, and taxonomy is loaded correctly
+        test_seq = 'TACGTAGGGAGCAAGCGTTGTCCGGAATTACTGGGTGTAAAGGGTGCGTAGGCGGGTATGCAAGTCATATGTGAAATACCGGGGCTCAACTCCGGGGCTGCATAAGAAACTGTATATCTTGAGTACAGGAGAGGTAAGCGGAATTCCTAG'
+        self.assertEqual(exp.feature_metadata['Taxon'][test_seq], 'k__Bacteria; p__Firmicutes; c__Clostridia; o__Clostridiales; f__Ruminococcaceae; g__; s__')
```
RNAer: the other advantage of using existing data is that we can assertequal read_qiime2 and read_amplicon so that the test can be done systematically, instead of just testing a single point.

calour/_testing.py
amnona: we need a new one since we need taxonomies and rep-seqs to also test the incorporation into the loaded experiment

calour/io.py
amnona: we cannot use the .open() because io.read() expects a file name and not a file.
so we need to unzip to a temp dir and then use io.read to load it

calour/_testing.py
RNAer: we can create the qza files for the existing data set

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Next version
 
+## Version 2019.5.2
```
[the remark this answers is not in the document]
amnona: fixed

calour/io.py
amnona: fixed

calour/io.py
amnona: endswith since qiime2 includes a random/hashed subfolder name in its zipfile
cannot use open() since biom.parse_table() uses the file.tell() method, which is supported for zipfile only in python 3.7 and not in 3.6 so parse_table fails.

calour/io.py
amnona: oops. fixed

calour/io.py
amnona: fixed

calour/io.py
amnona: addressed in prev. comment. only works in python 3.7

calour/io.py
amnona: there is no guarantee case will be preserved in taxonomy/rep seqs so better to convert. agree we lose info from the case/

calour/tests/test_io.py
amnona: you will create a single test dataset in a following PR :)

calour/io.py
amnona: yup. see for example:
a=pd.DataFrame({'a':[1,2,3]},index=[4,5,6])
b=pd.DataFrame({'aa':[5,6,4]},index=[6,5,4])
c=pd.concat([a,b],axis=1)
print(c)


   a  aa
4  1   4
5  2   6
6  3   5

amnona:
ok, addressed comments. can you merge?

calour/io.py
RNAer: this for loop can be replaced by list comprehension for concise and speed.

amnona: cool. fixed

amnona:
ok, now ready for merge :)

RNAer:
travis fail

amnona:
the travis errors are not related to the qiime2 changes:
2 error are in test_training
3rd error is in the test_save for test2.biom - again not related to the qiime2 changes (maybe biom changed?)
anyway let's merge the qiime2 PR and then address these problems.

amnona:
fixed the travis xvfd bug so now finally passing
@RNAer can u merge?

PR #159, opened 2019-08-12
Dev

maque4004:
add filter_features function


amnona:
looks good. thanks!
@RNAer  any comments?

PR #162, opened 2019-08-14
Improve q2 read

amnona:
fix when loading representative seqs and table contains a subset of them to use a join()


amnona:
@RNAer can you merge?

calour/io.py
```diff
@@ -177,12 +177,18 @@ def read_qiime2(fp, sample_metadata_file=None, rep_seq_file=None, taxonomy_file=
...
-            newexp.feature_metadata['_feature_id'] = rep_seqs
-            newexp.feature_metadata.set_index('_feature_id', inplace=True)
+            rseqs = []
+            rids = []
+            for cseq in skbio.read(rs_name, format='fasta'):
+                rseqs.append(str(cseq).upper())
+                rids.append(cseq.metadata['id'])
+            rep_seqs = pd.Series(data=rseqs, index=rids, name='_feature_id')
+            if not newexp.feature_metadata.index.equals(rep_seqs.index):
+                logger.info('Rep seqs hashes and table hashes are not equal. Using table hashes.')
```
RNAer: i am confused. if hash is not equal, how to match up? if you are using table hash anyway, why need to compare them first?

amnona: the problem arises when a user creates a biom table (and corresponding rep-seqs file) and then does some filtering (in qiime2) on the biom table.
then when he tries to load them into calour, there are more entries in the rep-seqs than in the biom-table (that was filtered). The old version just failed. This version takes the rep. seq for each hash in the biom table (and ignores the rep-seqs that don't have a hash)

PR #163, opened 2019-08-15
add figsize parameter to heatmap funciton

maque4004:
add figsize parameter to heatmap funciton
@amnona  Could you please check it out? Thanks!


amnona:
Cool. Thanks :)

amnona:
just need to update CHANGELOG.md (also with previous additions to API)

RNAer:
@amnona , let's add later. could you test if it works? @mortonjt mentioned to me that changing this only is not enough.

PR #168, opened 2019-11-27
Fix dsfdr and load repseqs from qiime2

amnona:
- Load repseqs file from qiime2 qza if supplied
- fix a bug in dsfdr when using a user supplied statist for dsfdr - we didn't take the abs() of the effect size so would only look for positive extreme effect size


calour/io.py
```diff
@@ -177,12 +177,18 @@ def read_qiime2(fp, sample_metadata_file=None, rep_seq_file=None, taxonomy_file=
...
+            rids = []
+            for cseq in skbio.read(rs_name, format='fasta'):
+                rseqs.append(str(cseq).upper())
+                rids.append(cseq.metadata['id'])
+            rep_seqs = pd.Series(data=rseqs, index=rids, name='_feature_id')
+            if not newexp.feature_metadata.index.equals(rep_seqs.index):
+                logger.info('Rep seqs hashes and table hashes are not equal. Using table hashes.')
+            newexp.feature_metadata.rename(columns={'_feature_id': '_hash'}, inplace=True)
+            newexp.feature_metadata = newexp.feature_metadata.join(other=rep_seqs, on='_hash', how='left')
+            newexp.feature_metadata.set_index('_feature_id', inplace=True, drop=False)
```
RNAer: what does this change diff?

calour/dsfdr.py
```diff
@@ -292,11 +291,13 @@ def dsfdr(data, labels, transform_type='rankdata', method='meandiff',
         # call the user-defined function of statistical test
         t = method(data, labels)
         tstat = t.copy()
+        t = np.abs(tstat)
```
RNAer: why need to be absolute value? could you add comment to explain?

calour/io.py
amnona: This was added so the sequences (and not the hashes) are the _feature_id and index
(added comment in file)

calour/dsfdr.py
amnona: We need abs() since dsFDR performs a two-sided test. added comment in file.

amnona:
comments addressed. @RNAer - ready for merge?

RNAer:
there are conflicts

amnona:
fixed the conflict (don't know why it created the conflict - same line in both versions....)
anyway, @RNAer  ready for merge?

PR #169, opened 2020-01-01
fix load qiime2 taxonomy

amnona:
bugfix when the table is a subset of the taxonomy. should join() instead of concat()


amnona:
@RNAer can u merge?

PR #170, opened 2020-01-02
fix qiime2 taxonomy

amnona:
the qiime2 taxonomy contains hashes and not sequences as key
chose to try both as there is no guarantee the taxonomy will be indexed on hash or taxonomy
@RNAer can u merge?


RNAer:
I don't understand your words. Pls write it clearly.

amnona:
When loading a Qiime2 taxonomy table, it is comprised of:
index, taxonomy, confidence
the index is not always the sequence, but rather depends on the method the table was generated (i.e. can be either sequence or hash).
In order to accommodate for both options (and since there is no guarantee this will not change in future qiime2 versions), this calour fix will try to compare the taxonomy file index to both hash (if present) and sequence of the loaded table.
@RNAer

PR #171, opened 2020-01-02
Fix sav_biom when feature metadata field not found

amnona:
If the metadata field (default is Taxonomy) is not found, save without it


amnona:
@RNAer small fix for save_biom - can you go over and merge?

PR #174, opened 2020-06-10
Fix pandas warn

amnona:
Fix a future behavior change in pandas:

smd = pd.concat([smd1, smd2], join='outer')
/Users/amnon/git/calour/calour/manipulation.py:260: FutureWarning: Sorting because non-concatenation axis is not aligned. A future version
of pandas will change to not sort by default.

To accept the future behavior, pass 'sort=False'.

To retain the current behavior and silence the warning, pass 'sort=True'.


amnona:
@RNAer can u merge?

RNAer:
you need to reset your local repo to get rid of the ugly local commit history

PR #175, opened 2020-06-17
Save fasta

amnona:
Add fasta header format options.
We need this since qiime2 picrust does not work with >NUM (conflicts with database name)
picrust2 standalone requires fasta with >SEQ in header
note by default it is compatible with the previous calour save_fasta


calour/io.py
```diff
@@ -799,26 +799,36 @@ def save_fasta(exp: Experiment, f, seqs=None):
     seqs : list of str sequences ('ACGT') or None, optional
         None (default) to save all sequences in exp, or list of sequences to only save these sequences.
         Note: sequences not in exp will not be saved
+    header_type: str, optional
+        The format for the per-sequence header in the output fasta file. options are:
+        'combined' (default): ">%d %s" % (NUM, SEQ)
+        'seq': ">%s" % SEQ
+        'taxonomy': ">Seq_%d %s" % (NUM, Taxonomy) (if available)
+        'num': ">Seq_"%d" % NUM
```
RNAer, line 807: why we need combined? isn't it redundant to have seq string on header? can't we just have 2 style:

1. ">seq_%d" % NUM
2. ">%s" % SEQ

And use additional taxonomy=True or False argument to add taxonomy string to either style or not.

amnona:
fixed. @RNAer can you merge?

PR #190, opened 2020-06-30
Fix heatmap autoscale

amnona:
By default, when supplying a norm, imshow uses vmin/vmax only for the first time it is initialized.
Since norm default is initialized in the function definition, it means vmin/vmax are only used the first time and then the previous range is used.
added a fix to renormalize each call


amnona:
see [here](https://matplotlib.org/api/_as_gen/matplotlib.pyplot.imshow.html) about vmin parameter ignored when norm is not None, and [here](https://matplotlib.org/api/_as_gen/matplotlib.colors.Normalize.html#matplotlib.colors.Normalize) about autoscale

Issue #193, opened 2020-07-02
plot for ampliconexperiment should work without dbbact-calour installed


amnona:
just print a warning the first time

amnona:
addressed in #197

PR #197, opened 2020-07-02
make database not installed warn instead of raise error

amnona:
Instead of raising an error, just warn the user and provide database interface installation instructions.

* Added installation, website and description to each database section in the calour.config file
* Added phenodb to config file


amnona:
address #193 
@RNAer can you merge and close the issue?

PR #198, opened 2020-07-03
Byebye skbio

amnona:
get rid of the skbio fasta iterator. replaced with our own (_iter_fasta() in calour.util)


calour/util.py
```diff
@@ -43,6 +43,44 @@
...
+    Fasta file must contain header line (starting with ">") and one or more sequence lines.
+
+    Parameters
+    ----------
+    file_name: str
+        name of the fasta file
+
+    Yields
+    ------
+    (header, sequence)
```
RNAer, line 57: delete this line

calour/util.py
```diff
@@ -43,6 +43,44 @@
...
+    header: str
+        the header line (without ">")
+    sequence: str
+        the sequence ('ACGT')
+    NOTE: header and sequence are stripped (no new line/spaces at beginning or end)
```
RNAer, line 62: change this 2 lines -> the sequence (eg 'ACGT'). Both header and sequence are whitespace stripped.

calour/util.py
```diff
@@ -43,6 +43,44 @@
 logger = getLogger(__name__)
 
 
+def _iter_fasta(file_name):
```
RNAer, line 46: not sure this should be private or public. but it definitely should be in io.py - could u move it? 

and could you rename `file_name` to `fp` to be consistent with other functions.

calour/util.py
```diff
@@ -43,6 +43,44 @@
...
+    '''Iterate over fasta file.
```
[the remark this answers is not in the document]
amnona: fixed

calour/util.py
amnona: fixed

calour/util.py
amnona: fixed

calour/util.py
amnona: file_name -> fp : fixed
agree more logical in io.py, but then we get cyclic import problem (io.py imports amplicon_experiment.py (for read() ), and amplicon_experiment.py needs _iter_fasta() for the filter_fasta() command). can think about moving filter_fasta() from amplicon_experiment.py, but it is specific for AmpliconExperiment (not a generic calour.Experiment since assumes features are sequences), so problem moving it to filtering.py.... Any idea how to solve?

RNAer: let me check the cyclic import issue.

PR #214, opened 2020-07-15
make skbio dependance optional

amnona:
Move skbio imports in transforming.py optional

### DIFF
--- a/calour/transforming.py
+++ b/calour/transforming.py
@@ -38,7 +38,6 @@ from copy import deepcopy
 import numpy as np
 from sklearn import preprocessing
 from skbio.stats.composition import clr, centralize as skbio_centralize
-from skbio.stats import subsample_counts
 
 from . import Experiment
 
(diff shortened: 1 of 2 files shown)


calour/transforming.py
```diff
@@ -302,6 +300,12 @@ def center_log_ratio(exp: Experiment, method=lambda matrix: matrix + 1, centrali
     skbio.stats.composition.clr
     skbio.stats.composition.centralize
     """
+    try:
+        from skbio.stats.composition import clr, centralize as skbio_centralize
+    except ModuleNotFoundError as err:
+        logger.error('scikit-bio must be installed in order to use this function')
+        raise err
```
RNAer, line 307: why you need to try and capture exception?? you are raising the err anyway!

amnona: In order to make better instructions on why it's failing - scikit-bio is not installed...
don't think it complicates too much but not critical - if you feel strongly about it i can delete...

RNAer: `ModuleNotFoundError: No module named 'dafasdf'` looks very clear to me. no need for extra code

amnona:
ready for merge :)

PR #215, opened 2020-07-15
Heatmap nan

amnona:
Add an option to specify a different color for the colormap.bad_color().
Used to be hard-set in the heatmap() code to 'black'. But sometimes we want a different color for the nans.
While we could use the bad_color set in the colormap parameter, problem is the default bad_color() (at least for viridis) is 'white', which does not look good (since all the 0s are log transformed to nans)

also updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # calour changelog
 
+## Version 2020.???
+Incompatible changes:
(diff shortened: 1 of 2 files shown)


amnona:
@RNAer small change can you merge?

amnona:
ready for merge :)

PR #220, opened 2020-07-20
fix cluster_features() bug

RNAer:
cluster_feature() shouldn't transform the data table.

### DIFF
--- a/calour/sorting.py
+++ b/calour/sorting.py
@@ -36,6 +36,7 @@ from scipy import cluster, spatial
 from . import Experiment
 from .filtering import _filter_ids
 from .transforming import log_n, standardize
+from .manipulation import chain
 from .util import _argsort
 from ._doc import ds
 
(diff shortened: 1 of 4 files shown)


calour/sorting.py
```diff
@@ -132,8 +133,7 @@ def cluster_data(exp: Experiment, transform=None, axis=1,
         data = exp.get_data(sparse=False)
     else:
         logger.debug('transforming data using %r' % transform)
-        newexp = deepcopy(exp)
-        data = transform(newexp, **kwargs).get_data(sparse=False)
+        data = transform(exp, **kwargs, inplace=False).get_data(sparse=False)
```
amnona, line 136: i think you need to add copy=True,
otherwise, if the experiment is already not sparse, will return the reference to the data instead of copy it (and then will change in place...)

RNAer:
I don’t think so. It Is copied because inplace is false

PR #228, opened 2020-09-21
add value to Experiment.iterate()

amnona:
So now exp.iterate() yields the value and the experiment subset.
For example:
exp.iterate('gender')
yields
'male', Experiment (with all male samples)
'female', Experiment (with all female samples)


amnona:
@RNAer  ready for merge :)

PR #230, opened 2020-09-22
fix debug message in delete sequence from annotation

amnona:
fix the debugging message in right_menu_remove_feature
(called when right clicking on a feature and selecting remove from annotation)


amnona:
@RNAer ready for merge

PR #235, opened 2020-10-14
Paired testing for diff. abundance

amnona:
Added diff_abundance_paired() method to test for differential abundance in pairs of samples
Fixed dsfdr (and diff_abundance() to accept shuffler, a function that can shuffle pairs instead of the default all samples.
Added "direction" transform to diff_abundance_paired() to enable direction binary testing
Added unit testing for the new function


amnona:
@RNAer-  a new and shiny paired diff abundance function
The "direction" transform is a bit less elegant, but didn't find a nicer way to implement.
comments? merge?

PR #240, opened 2020-10-22
fix bar_label_kwargs so rotation works

amnona:
Add option for "rotation" parameter in barx_label_kwargs / bary_label_kwargs
(previous implementation it was not possible to change rotation since provided in the function explicitly)


amnona:
@RNAer merge?

RNAer:
`xticklabel_kwargs=None, yticklabel_kwargs=None, ` parameters can already do that

amnona:
How can you rotate the xticklabels? it had rotation hard coded so always 0 if xticklabels. and if you try to supply it in the xticklabel_kwargs, you get parameter defined twice...

RNAer:
it's not hard coded - it's just a default and will be updated when you provide it in xticklabel_kwargs.

PR #241, opened 2020-11-04
Paired groups

amnona:
* Add the option to have more than one sample per paired-group value. permutations are still down within each paired group.
For example, now can have for an individual 3 samples before treatment, 2 samples after, and permutations will be on these 5 samples
* changed the name of the "direction" paired transform to "pair_rank", as it does ranking on all the samples in each group (similar to direction if only one sample in each paired-group value)
* bugfix: when using the "pair_rank" transform (used to be "direction"), we modified the actual experiment data instead of a copy. fixed.


amnona:
@RNAer merge?

amnona:
@RNAer ping :)

PR #244, opened 2020-11-23
fix bug when not using all samples in paired test

amnona:
Fixed a bug where diff_abundance_paired would fail if the 2 values (val1, val2) did not cover all the samples.
This happened since we set up the _pair_shuffler in diff_abundance_paired() but did the sample filtering (according to val1,val2) later in diff_abundance()
So the fix now does the sample filtering in diff_abundance_paired()


amnona:
@RNAer small bug fix. can you merge?

PR #245, opened 2020-11-26
Byebye skbio

amnona:
make skbio dependency optional (only when using the dendrogram tree)
added to test optional dependency since we have a dendrogram unit test


amnona:
@RNAer for some reason travis is not running the tests. any idea why?
also, can you go over and comment/merge?

setup.py
```diff
@@ -65,11 +65,12 @@
...
           'biom-format',
           'statsmodels',
           'ipywidgets',
           'docrep'],
-      extras_require={'test': ["nose", "pep8", "flake8"],
+      extras_require={'test': ["nose", "pep8", "flake8", 'scikit-bio >= 0.5.1'],
                       'coverage': ["coveralls"],
-                      'doc': ["Sphinx >= 1.4", "sphinx-autodoc-typehints", "nbsphinx"]})
+                      'doc': ["Sphinx >= 1.4", "sphinx-autodoc-typehints", "nbsphinx"],
+                      'dendrogram': ['scikit-bio >= 0.5.1']
```
RNAer, line 75: you need to update travis setting to add `dendrogram` installation? to make sure that  the test code for dendrogram and plot_tree is tested

amnona: I think it already contains the skbio requirement (in ci/conda_requirements.txt) so all tests should run ok.
But github is still not running the travis on this PR. any idea why?

amnona: I see it passed on travis:
https://travis-ci.org/github/amnona/calour/builds/746442457
but don't see it here... what's going on?

PR #252, opened 2021-04-07
Dbbact version

amnona:
Moved dbbact-calour version numbers to semantic versioning (i.e. '1.4.1')
Added support for parsing these version numbers when getting the minimal database version from the calour.config file.
If database version is lower than the recommended version in the calour.config, it displays a warning to the user
Also updated calour.config for the new semantic versioning minimal recommended dbbact-calour version (1.4)


amnona:
@RNAer fixed conflicts. can u merge?

PR #256, opened 2021-07-27
feat: support https on the exported heatmap

amnona:
dbbact now works https
converting exported heatmap query links to https (since we get mixed content warning if query using http from an https page)


amnona:
@RNAer can you merge?

PR #272, opened 2022-07-23
fix join_metadata() and version bump

amnona:
the default should be 0 rather than 's' since the conversion from 's'->0 is performed if the parameter is supplied (not on the default)


amnona:
@RNAer can you merge?

PR #274, opened 2022-07-25
fix heatmap() inplace colormap modification depracation warning

amnona:
copy the colormap so we don't modify in place (otherwise we get a matplotlib deprecation warning)


amnona:
@RNAer can you merge?

RNAer:
conflict to resolve

amnona:
fixed the confilct.
In general, we always get conflicts if we do different branches for different fixes, and write the changelog for each fix.
what is the best way to do it?

RNAer:
hmm, good question. not sure on top of my head...

PR #275, opened 2022-07-25
fix _read_metadata() to show as warning/info only the summary 

amnona:
Instead of giving a warning with all the missing sample data/metadata, show only a summary and first 5 samples without data/metadata.
The full list of problematic samples is available as a debug() level message.


amnona:
@RNAer  can you merge?

PR #276, opened 2022-07-25
remove normaliztion check warning from filter_by_data()

amnona:
To prevent the warning when reading an experiment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ## Version 2022.7.1
 Bug Fixes:
 * Fix join_metadata() to use axis='s' by default
+* Remove normalization check from filter_by_data()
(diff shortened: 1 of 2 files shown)


RNAer:
why removing this? a friendly warning doesn't hurt

amnona:
Since we filter (for min_reads) prior to normalization, you get this warning every time you load an experiment (for example using read_amplicon). It is confusing to the user.
In general, sometimes you want to filter by data even without normalization (for example when looking at metabolomics, which is absolute abundance and therefore no need to normalize), so this warning is not needed.
Also, since you must specify the normalization, the user is aware of the fact if his data is normalized or not

amnona:
resolved conflict again :)

PR #277, opened 2022-07-26
add info after loading an experiment

amnona:
Info about the loaded experiment (how many samples/features)


amnona:
@RNAer can you merge?

RNAer:
better to use `exp.shape`?

amnona:
@RNAer addressed - c u merge?

RNAer:
not a big deal. for simplicity, you could just use `exp.shape` instead of `exp.shape[0], exp.shape[1]`.

PR #278, opened 2022-07-27
Better diff_abundance_enrichment() plot

amnona:
* Add option to specify the bary_labels style kwargs and the within-bar (num enriched experiments) text style kwargs
* Add warning if (usually due to bad background script in dbBact) term does not have total_experiments count


amnona:
@RNAer can you merge? once you merge i can update the changelog so will be ready for next time without the conflict

PR #279, opened 2022-07-27
Update reporting to diff abundance and filtering

amnona:
* Add better reporting to filter_by_data() and various diff abundance methods
* add sanity checks to diff abundance methods
* remove old parameter doc


amnona:
@RNAer can you merge?

PR #280, opened 2022-07-29
enable sep=XX parameter in _read_metadata()

amnona:
so can specify a different separator for the metadata when reading an experiment


amnona:
@RNAer can u merge?

PR #63, opened 2017-10-06
WIP add a new module for machine learning functionalities

RNAer:

PR #295, opened 2024-05-01
Mpl37fix (#63)

amnona:
* add some more type hints

* create hashes if not provided in new experiment

* bugfix update remove lines for mpl3.7

* update changelog


amnona:
@RNAer can you merge?

PR #309, opened 2024-08-30
add shuffler option to analysis.correlation()

amnona:
So can use it for paired correlation testing
Also improve data validation in the correlation() function


RNAer:
Looks good to merge. compared to the previous code, it basically opens api for user to choose the permute method?

could you resolve the conflict? or we should review and merge the PR before this one?

amnona:
Yup, this enables choosing the permute function also for the correlation() method (similar to the diff_abundance). Useful when doing paired correlation tests.

I resolved the changelog conflict. feel free to merge :)

